PR #15, opened 2022-05-05
added dependabot 

acn-sbuad:
- Ignoring major updates. Do we want to continue with this pratcice?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,9 +3,6 @@ updates:
   ## Register
   - directory: "src/"
     package-ecosystem: "nuget"
-    ignore:
-      - dependency-name: "*"
-        update-types: ["version-update:semver-major"]
     schedule:
       interval: "weekly"
       day: "thursday"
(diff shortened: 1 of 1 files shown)


SandGrainOne:
> Ignoring major updates. Do we want to continue with this pratcice?
Major versions will often have breaking changes, but being reminded of them from dependabot could in itself have some value.

acn-sbuad:
> > Ignoring major updates. Do we want to continue with this pratcice?
> 
> Major versions will often have breaking changes, but being reminded of them from dependabot could in itself have some value.


With this config we would get a notification about the major updates, given there is minor updates as well. If there are no minor updates we will not be notified about the majors either. 
We could have a job that runs more seldom or at the same frequency where only the major-updates are included. What do you think about this approach ?

SandGrainOne:
We can allow dependabot to make pull requests for all updates, including major. We just need to take care not to merge updates to major versions without a more thorough check of what the update contains.

SandGrainOne (approved):

Issue #9, opened 2022-04-08
Remove the term Platform from all projects and namespaces


acn-sbuad:
## Description 
`Altinn.Platform.Register` is repeated many places. This should be simplified to be `Altinn.Register`.

## Specification 
Folder structure where application projects are in a src folder and test projects are in a test folder.

src/Altinn.Register/Altinn.Register.csproj
test/Altinn.Register.Tests/Altinn.Register.Tests.csproj

Most of the structure is in place already, but the src/Register folder should be renamed to src/Altinn.Register.

## Tasks
- [x] Remove `Platform` from all namespaces
- [x] Restructure repository
- [x] Update github actions as needed
- [x] Update build pipelines in Azure DevOps

PR #48, opened 2022-08-04
Simplified paths and namespaces

acn-sbuad:
## Description

- Register code directly unde src/
- Test code directly under test/
- Removed Altinn.Platform prefix in folders and namespaces. 
- Updated all relevant references
- TODO: Update pipelines in DevOps

## Related Issue(s)
- #9

## Verification
- [x] All tests run green


SandGrainOne:
I think we should keep `Altinn` like we did in Notification. I also think we should keep the project folders. I want us to (eventually) split the application project into multiple projects in the same way as we did for Notification.

PR #100, opened 2023-01-15
Added Getpartylist in register

acn-dgopa:
Added endpoint for retreiving partylist for a list of partyids

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


src/Controllers/PartiesController.cs
```diff
@@ -98,6 +99,35 @@ public async Task<ActionResult<Party>> PostPartyLookup([FromBody] PartyLookup pa
...
+        /// <param name="partyIds">List of partyIds for parties to retrieve.</param>
+        /// <returns>List of parties based on the partyIds.</returns>
+        [HttpPost("partylist")]
+        [Consumes("application/json")]
+        [Produces("application/json")]
+        public async Task<ActionResult<List<Party>>> GetPartyListForPartyIds([FromBody] List<int> partyIds)
+        {
+            List<Party> parties;
+
+            try
```
SandGrainOne, line 114: Remove the try-catch. The middleware will produce a response with status code 500 if there is an exception. A test shold be able to verify this.

src/Services/Implementation/PartiesWrapper.cs
```diff
@@ -17,6 +19,7 @@ namespace Altinn.Register.Services.Implementation
     /// <summary>
     /// The parties wrapper
     /// </summary>
+    [ExcludeFromCodeCoverage]
```
SandGrainOne, line 22: It's possible to create tests by mocking a HttpMessageHandler. We have examples of this in Profile. Look for [DelegatingHandlerStub](https://github.com/Altinn/altinn-profile/blob/42308b53b8e2949d21e5e7e434e03b76a4db31e0/test/Altinn.Profile.Tests/Mocks/DelegatingHandlerStub.cs) and how we use it.

src/Services/Implementation/PartiesWrapper.cs
```diff
@@ -110,5 +113,35 @@ public async Task<int> LookupPartyIdBySSNOrOrgNo(string lookupValue)
 
             return -1;
         }
+
+        /// <inheritdoc />
+        public async Task<List<Party>> GetPartyList(List<int> partyIds)
+        {
+            try
```
SandGrainOne, line 120: Remove the try catch. The middleware should handle the exception logging.

SandGrainOne (changes requested):
A few things I would like to have changed

test/Utils/TestDataLoader.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            string path = $"../Testdata/{typeof(T).Name}/{id}.json";
+            string fileContent = await File.ReadAllTextAsync(GetPath(id));
+            T data = JsonSerializer.Deserialize<T>(fileContent);
+            return data;
+        }
+
+        private static string GetPath(string id)
+        {
+            string? unitTestFolder = Path.GetDirectoryName(new Uri(typeof(PartiesControllerTests).Assembly.Location).LocalPath);
```
SandGrainOne, line 21: The questionmark on `string?` is unnecessary.

test/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        [Fact]
+        public async Task GetPartyListForPartyIds_InvalidInput_Notfound()
+        {
+            // Arrange
+            List<int> partyIds = new List<int>();
+            partyIds.Add(1);
+            partyIds.Add(2);
+
+            HttpRequestMessage sblRequest = null;
+            DelegatingHandlerStub messageHandler = new(async (request, token) =>
```
SandGrainOne, line 84: Should we consider making an assert ot two on the request being sent to SBL Bridge? Alternatively remove the variable.

test/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            HttpRequestMessage testRequest = new HttpRequestMessage(HttpMethod.Post, "/register/api/v1/parties/partylist") { Content = requestBody };
+            testRequest.Headers.Add("PlatformAccessToken", PrincipalUtil.GetAccessToken("ttd", "unittest"));
+
+            // Act
+            HttpResponseMessage response = await client.SendAsync(testRequest);
+
+            // Assert
+            Assert.Equal(response.RequestMessage.RequestUri, testRequest.RequestUri);
+            Assert.Equal(response.RequestMessage.Content, testRequest.Content);
```
SandGrainOne, line 68: I don't trust these asserts to actually test anything of value. I suspect that they are comparing one object with itself. In other words, what we have here are two variables/pointers pointing to the same address in memory.

acn-dgopa: i have now udpated the asserts similar to profile component. I hope the udpated asserts makes sense :)

PR #115, opened 2023-04-17
Changed policy setup to not require jwt cookie when not needed.

TheTechArch:
It removed check of JWT End User token. This was not needed or used for anything for these two methods. 

Updated test to verify this.

### DIFF
--- a/src/Controllers/PartiesController.cs
+++ b/src/Controllers/PartiesController.cs
@@ -19,6 +19,7 @@ namespace Altinn.Register.Controllers
     /// <summary>
     /// The parties controller provides access to party information in the SBL Register component.
     /// </summary>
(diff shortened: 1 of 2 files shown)


src/Controllers/PartiesController.cs
```diff
@@ -19,8 +19,6 @@ namespace Altinn.Register.Controllers
     /// <summary>
     /// The parties controller provides access to party information in the SBL Register component.
     /// </summary>
-    [Authorize]
-    [Authorize(Policy = "PlatformAccess")]
```
SandGrainOne, line 23: Could we have kept PlatformAccess token here on the top?

TheTechArch: Fixed

PR #156, opened 2023-12-13
Fixed attempts to 3. Verified ok with JVS

TheTechArch:
Changed the number of attempts to 3. Verified OK with JVS/EM


SandGrainOne (approved):
Good work! 🥇

PR #169, opened 2024-02-01
Add party name lookup endpoint with internal admin access scope requirement

elsand:
## Description
This adds an new authorization policy implementing a requirement akin to [PublishScopeOrAccessTokenRequirement](https://github.com/Altinn/altinn-events/blob/main/src/Events/Authorization/PublishScopeOrAccessTokenRequirement.cs), allowing PartiesController actions to be accessible with either an AccessToken or with an Altinn Token Exchange-issued token with a particular internal scope.

## Related Issue(s)
- https://github.com/digdir/dialogporten/issues/321

## Verification
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green

## Documentation


src/Program.cs
```diff
@@ -213,6 +216,7 @@ void ConfigureServices(IServiceCollection services, IConfiguration config)
         options.AddPolicy("PlatformAccess", policy => policy.Requirements.Add(new AccessTokenRequirement()));
         options.AddPolicy("AuthorizationLevel2", policy =>
             policy.RequireClaim(AltinnCore.Authentication.Constants.AltinnCoreClaimTypes.AuthenticationLevel, "2", "3", "4"));
+        options.AddPolicy("InternalOrPlatformAccess", policy => policy.Requirements.Add(new InternalScopeOrAccessTokenRequirement("altinn:register/partylookup.admin")));
```
SandGrainOne, line 219: Probably need to add ScopeAccessHandler to services.

src/Services/Implementation/PartiesWrapper.cs
```diff
@@ -159,4 +164,64 @@ public async Task<List<Party>> GetPartyListByUuid(List<Guid> partyUuids)
         _logger.LogError("Getting parties information from bridge failed with {StatusCode}", response.StatusCode);
         return new List<Party>();
     }
+
+    /// <inheritdoc />
+    public async Task<PartyNamesLookupResult> LookupPartyNames(PartyNamesLookup partyNamesLookup)
+    {
+        var partyNames = new ConcurrentBag<PartyName>();
+        var tasks = new List<Task>();
+        var semaphore = new SemaphoreSlim(_concurrentNameLookups);
```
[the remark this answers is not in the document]
elsand: The semaphore is used to to ensure no more than `_concurrentNameLookups` concurrent Bridge lookups, regardless of the size of `partyNamesLookup.Parties`.

src/Program.cs
elsand: Ah yes, this was the missing piece. Now auth works locally :)

src/Controllers/PartiesController.cs
```diff
@@ -147,6 +147,30 @@ public async Task<ActionResult<Party>> PostPartyLookup([FromBody] PartyLookup pa
 
             return Ok(party);
         }
+        
+        /// <summary>
+        /// Perform a name lookup for the list of parties for the provided ids.
+        /// </summary>
+        /// <param name="partyNamesLookup">A list of lookup criteria. For each criteria, one and only one of the properties must be a valid value.</param>
+        /// <returns>The identified party names for the corresponding identifiers.</returns>
+        [ValidateModelState]
```
SandGrainOne, line 156: Was there a plan to add validation to the model? Should probably have asked about that a bit sooner.

src/Services/Implementation/PartiesWrapper.cs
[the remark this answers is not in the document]
SandGrainOne: Not sure how this works, but if we get 2 concurrent requests to do this lookup would we have 2 SemaphoreSlim instances that would allow for a total of 20 concurrent lookups? 

I'm asking simply to understand. I don't mind the current implementation. Such a scenario would be very rare I assume.

test/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -116,4 +116,140 @@ public async Task GetPartyListForPartyIds_InvalidInput_Notfound()
...
+        testRequest.Headers.Add("PlatformAccessToken", PrincipalUtil.GetAccessToken("ttd", "unittest"));
+
+        // Act
+        HttpResponseMessage response = await client.SendAsync(testRequest);
+        string responseContent = await response.Content.ReadAsStringAsync();
+
+        PartyNamesLookupResult actualResult = JsonSerializer.Deserialize<PartyNamesLookupResult>(
+            responseContent, options);
+
+        PartyNamesLookupResult actualResultFromCache = JsonSerializer.Deserialize<PartyNamesLookupResult>(
```
SandGrainOne, line 178: One idea here is to look in the actual cache. We have access to the MemoryCache being used: `_webApplicationFactorySetup.MemoryCache`

SandGrainOne: This also allows us to add an element to the cache before running the test in order to test the use of the cache.

src/Controllers/PartiesController.cs
```diff
@@ -147,6 +147,30 @@ public async Task<ActionResult<Party>> PostPartyLookup([FromBody] PartyLookup pa
...
+        [Consumes("application/json")]
+        [ProducesResponseType(400)]
+        [ProducesResponseType(404)]
+        [ProducesResponseType(200)]
+        [Produces("application/json")]
+        public async Task<ActionResult<PartyNamesLookupResult>> PostPartyNamesLookup([FromBody] PartyNamesLookup partyNamesLookup)
+        {
+            PartyNamesLookupResult partyNamesLookupResult = await _partiesWrapper.LookupPartyNames(partyNamesLookup);
+
+            if (partyNamesLookupResult == null)
```
SandGrainOne, line 167: Looks like it's impossible for `partyNamesLookupResult` to be null. Can't see any call chain where null is returned.

SandGrainOne (approved):

src/Controllers/PartiesController.cs
elsand: This utilizes the existing validation for `PartyLookup`, but I suppose one could add validation to ensure that `PartyNamesLookup.Parties`, ie the `List<PartyLookup>`, is an array/not null. Passing in `{ "parties": null }` will result in a NullReferenceException. Should we create a 1.4.1 in Altinn.Platform.Models where a validator is implemented?

test/IntegrationTests/PartiesControllerTests.cs
elsand: Yes, but that would require the test code to know how to construct the cache key, which seems to me ought to be information/logic internal to the system-under-test.

src/Controllers/PartiesController.cs
elsand: Removed the if clause.

src/Controllers/PartiesController.cs
SandGrainOne: Nah, this is fine. I didn't notice that the new models used PartyLookup to be honest.

src/Services/Implementation/PartiesWrapper.cs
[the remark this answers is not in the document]
elsand: That is correct. Each request will get its own instance of SemaphoreSlim(10), so 2 concurrent request to this endpoint (with >= 10 input parties) might cause up to 2*10 concurrent requests to sblbridge. We can utilize a static instance of SemaphoreSlim make effectively make this an application-wide (ie. per pod) limit, which is probably a safer approach. I'll change this.

PR #180, opened 2024-03-15
feat: reduce affordances provided by `IParties`

Alxandr:
This is in preparation for supporting a local development mode of register with reduced features but no dependency on Altinn 2. This commit creates a new Altinn.Register.Core project and moves and renames the `IParties` interface to it. It also updates the interface with more idiomatic naming of methods and ensure that they all take in `CancellationToken`s. Lastly, it reduced the affordances provided by the interface by limiting the size of it's outbound contract while also relaxing the input constraints.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


test/Altinn.Register.Tests/Utils/AsyncList.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Altinn.Register.Tests.Utils;
+
+internal class AsyncList<T>
+    : IAsyncEnumerable<T>
+    , IEnumerable<T>
```
SandGrainOne, line 10: Fix the warnings:
```suggestion
    : IAsyncEnumerable<T>, IEnumerable<T>
```

test/Altinn.Register.Tests/UnitTests/ListExtensionsTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using System.Collections.Generic;
+using Altinn.Register.Extensions;
+using Xunit;
+
+namespace Altinn.Register.Tests.UnitTests;
+
+public class ListExtensionsTests
+{
+    public static TheoryData<List<int>, int, List<int>> SwapRemoveCases => new()
```
SandGrainOne, line 12: All tests work on integers, but the use of the type is on Tasks. Would it be interesting to add tests using Tasks?

I assume List.IndexOf will find correct index by comparing the reference value and not the Task, but would it be interesting to prove that with a test?

src/Altinn.Register/Services/Implementation/PartiesWrapper.cs
```diff
@@ -101,123 +146,123 @@ public async Task<Party> LookupPartyBySSNOrOrgNo(string lookupValue)
...
     {
-        if (_memoryCache.TryGetValue(cacheKey, out string partyName))
+        if (_memoryCache.TryGetValue(cacheKey, out string? partyName) 
+            && partyName is not null)
         {
             return partyName;
         }
 
-        Party party = await LookupPartyBySSNOrOrgNo(lookupValue);
+        // limit the concurrent calls to spl bridge
```
SandGrainOne, line 254:
```suggestion
        // limit the concurrent calls to SBL Bridge
```

SandGrainOne (approved):

test/Altinn.Register.Tests/UnitTests/ListExtensionsTests.cs
Alxandr: Whether they are tasks or not should not matter - and debugging/reading errors where you have 2 tasks that are not the same would be really annoying. But I can add some tests with `string` which is also a reference type (ie, the list just contains a pointer - same as with Task).

Alxandr: I added some string tests.

PR #255, opened 2024-07-17
chore: cleanup Altinn.Platform.Models

Alxandr:
feat: enable nullable reference types in Altinn.Platform.Models
  BREAKING-CHANGE: This change enables nullable reference types in
  Altinn.Platform.Models. All reference types are set as nullable by
  default and is likely to cause consumers of this library to get
  compiler warnings where they previously did not. Eventually, some of
  these properties will be set as non-nullable. We do not consider this
  a breaking change.

feat: make lists in Altinn.Platform.Models `IReadOnlyList<T>`
  BREAKING-CHANGE: This change makes all lists in Altinn.Platform.Models
  `IReadOnlyList<T>` instead of `List<T>`. This is to enable us to
  change the implementation of these lists in the future without
  breaking consumers of the library. The list properties are still
  writable, so the entire list can be replaced if for some reason that
  is needed.

feat: make most types record
  Most types in Altinn.Platform.Models are now records. This mostly
  provides better `ToString` implementations. This affects how equality
  is determined for these types, but we do not consider this a breaking
  change.

chore: mark `LanguageType` in `ProfileSettingPreference` obsolte
  BREAKING-CHANGE: The `LanguageType` property in
  `ProfileSettingPreference` is marked obsolete. It is a set-only
  property which just proxies to the `Language` property and is only
  present for historical reasons. It will eventually be removed.

chore: make Altinn.Platform.Models target .net 8.0
  BREAKING-CHANGE: Altinn.Platform.Models now targets .net 8.0. This
  means that consumers of this library must also target .net 8.0 or
  later.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,7 +8,6 @@
   <ItemGroup>
     <PackageVersion Include="Altinn.Common.AccessToken" Version="4.4.3" />
     <PackageVersion Include="Altinn.Common.PEP" Version="4.0.0" />
-    <PackageVersion Include="Altinn.Platform.Models" Version="1.6.1" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.1" />
     <PackageVersion Include="Azure.Identity" Version="1.12.0" />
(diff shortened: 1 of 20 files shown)


TheTechArch (approved):
LGTM

PR #290, opened 2024-09-05
test: fix pathological worst case of serial tests

Alxandr:
When tests using DB were run serially, they would not reuse the same database server, and instead each create their own container, leading to substantial overhead. This was fixed by delaying cleanup of the container for a few milliseconds after each test, to allow the next test to reuse the same container.

## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation


TheTechArch (approved):
LGTM.  BIDNUEATBIITT

PR #302, opened 2024-09-16
Add criteria to split name into first, middle, and last.

Ahmed-Ghanam:
## Description

Extend search criteria to determine  whether a person's first, middle, and last names should be included in the search results. Additionally, add properties for first, middle, and last names to the search result model.

## Related Issue(s)
- [593 Names lookup in Register that Notifications can use in keyword replacement](https://github.com/Altinn/altinn-notifications/issues/593)

## Verification
- [x] Manual testing done (required)
- [x] All tests run green

## Documentation


src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PartyName.cs
```diff
@@ -20,8 +20,29 @@ public record PartyName
...
     /// </summary>
     [JsonPropertyName("name")]
     public string? Name { get; set; }
+
+    /// <summary>
+    /// Gets or sets the first name for this result.
+    /// </summary>
+    [JsonPropertyName("firstName")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? FirstName { get; set; }
```
Alxandr, line 33: I think these 3 fields should probably be put inside a sub-object. The reason for that is currently, even when you are including the name components, if for instance middle name is `null`, the field will not be included. But if you put them in a sub-object (`nameComponents` for instance), the whole thing will be `null` or not at once.

Alxandr (changes requested):

src/Altinn.Register/src/Altinn.Register/Clients/PartiesClient.cs
```diff
@@ -267,8 +270,19 @@ private async Task<PartyName> ProcessPartyLookupAsync(PartyLookup partyLookup, C
...
+            };
+
+            if (shouldSplitPersonName && party.Person != null)
+            {
+                partyName.LastName = party.Person.LastName;
+                partyName.FirstName = party.Person.FirstName;
+                partyName.MiddleName = party.Person.MiddleName;
+            }
+
+            _memoryCache.Set(cacheKey, partyName, new TimeSpan(0, _cacheTimeoutForPartyNames, 0));
```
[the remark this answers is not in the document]
Ahmed-Ghanam: I have update the cache key so it takes into consideration the component to include.

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PartyLookup.cs
```diff
@@ -24,6 +24,15 @@ public record PartyLookup
...
     public string? OrgNo { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the person's name should be split into first, middle, and last name components.
+    /// </summary>
+    /// <value>
+    ///   <c>true</c> if person's name should be split into first, middle, and last name components; otherwise, <c>false</c>.
+    /// </value>
+    [JsonPropertyName("splitPersonName")]
+    public bool SplitPersonName { get; set; } = false;
```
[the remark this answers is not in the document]
Ahmed-Ghanam: The following PartyComponentOptions is implemented:
```
/// <summary>
/// Represents options for including different components of a party.
/// </summary>
[Flags]
public enum PartyComponentOptions
{
    /// <summary>
    /// No extra components are included.
    /// </summary>
    None = 0,

    /// <summary>
    /// Include the party's first, middle, and last name.
    /// </summary>
    NameComponents = 1 << 0,
}

```

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PartyName.cs
Ahmed-Ghanam: These three fields are now wrapped in the new type `PersonNameComponents`

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PersonNameComponents.cs
```diff
@@ -0,0 +1,30 @@
...
+/// <summary>
+/// Represents the components of a person's name.
+/// </summary>
+public record PersonNameComponents
+{
+    /// <summary>
+    /// Gets or sets the first name.
+    /// </summary>
+    [JsonPropertyName("firstName")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
```
Alxandr, line 14: These should probably not have `JsonIgnore`

src/Altinn.Register/src/Altinn.Register/Clients/PartiesClient.cs
```diff
@@ -232,22 +235,27 @@ public IAsyncEnumerable<PartyName> LookupPartyNames(IEnumerable<PartyLookup> loo
...
     {
         Debug.Assert(!string.IsNullOrEmpty(partyLookup.Ssn) || !string.IsNullOrEmpty(partyLookup.OrgNo));
+
         string lookupValue = !string.IsNullOrEmpty(partyLookup.Ssn) ? partyLookup.Ssn : partyLookup.OrgNo!;
-        string cacheKey = $"n:{lookupValue}";
-        string? partyName = await GetOrAddPartyNameToCacheAsync(lookupValue, cacheKey, cancellationToken);
+
+        bool includeNameComponents = !string.IsNullOrEmpty(partyLookup.Ssn) && partyLookup.IncludeComponents == PartyComponentOptions.NameComponents;
+
+        string cacheKey = string.IsNullOrEmpty(partyLookup.Ssn) ? $"n:{lookupValue}" : $"n:{lookupValue}-c:{includeNameComponents}";
```
Alxandr, line 243: Why are you caching results differently based on whether or not you want the name component? That seems like throwing away good data for no apparent reason?

src/Altinn.Register/src/Altinn.Register/Clients/PartiesClient.cs
```diff
@@ -232,22 +235,27 @@ public IAsyncEnumerable<PartyName> LookupPartyNames(IEnumerable<PartyLookup> loo
     private async Task<PartyName> ProcessPartyLookupAsync(PartyLookup partyLookup, CancellationToken cancellationToken)
     {
         Debug.Assert(!string.IsNullOrEmpty(partyLookup.Ssn) || !string.IsNullOrEmpty(partyLookup.OrgNo));
+
         string lookupValue = !string.IsNullOrEmpty(partyLookup.Ssn) ? partyLookup.Ssn : partyLookup.OrgNo!;
-        string cacheKey = $"n:{lookupValue}";
-        string? partyName = await GetOrAddPartyNameToCacheAsync(lookupValue, cacheKey, cancellationToken);
+
+        bool includeNameComponents = !string.IsNullOrEmpty(partyLookup.Ssn) && partyLookup.IncludeComponents == PartyComponentOptions.NameComponents;
```
Alxandr, line 241: Don't use equality checks with flags enums like this, instead use `HasFlags`. Also, you don't need to care if `Ssn` is set.

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PersonNameComponents.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Altinn.Platform.Register.Models;
+
+/// <summary>
+/// Represents the components of a person's name.
+/// </summary>
+public record PersonNameComponents
```
Alxandr, line 8: Seeing as this is a new record, probably just make it immutable by replacing all of the `set;` with `init;`.

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +400,115 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
+        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+        actualResult.Should().BeEquivalentTo(expectedResult);
+        actualResultFromCache.Should().BeEquivalentTo(expectedResult);
+    }
+
+    /// <summary>
+    /// Provides test data for parameterized tests.
+    /// </summary>
+    public static TheoryData<PartyLookup> GetPartyLookupTestData()
```
Alxandr, line 506: Xunit prefers theory data being (xunit)-serializable. It would be better to make this a `TheoryData<string, PartyComponentOptions>`, then build the actual `PartyLookup` in the test method.

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PersonNameComponents.cs
Ahmed-Ghanam: Keeping the `JsonIgnore` attribute on the properties within `PersonNameComponents` ensures consistent behavior across different contexts where `PersonNameComponents` might be used.

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +400,118 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
+
+        HttpRequestMessage testRequest = new(HttpMethod.Post, "/register/api/v1/parties/nameslookup") { Content = requestBody };
+
+        testRequest.Headers.Add("PlatformAccessToken", PrincipalUtil.GetAccessToken("ttd", "unittest"));
+
+        // Act
+        HttpResponseMessage response = await client.SendAsync(testRequest);
+        string responseContent = await response.Content.ReadAsStringAsync();
+
+        PartyNamesLookupResult actualResult = JsonSerializer.Deserialize<PartyNamesLookupResult>(responseContent, options);
```
Alxandr, line 490: You're deserializing the same content twice? Why? And what makes one of these "from cache", when it's the same content?

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PartyLookup.cs
```diff
@@ -24,6 +26,12 @@ public record PartyLookup
     [RegularExpression("^[0-9]{9}$", ErrorMessage = "Value needs to be exactly 9 digits.")]
     public string? OrgNo { get; set; }
 
+    /// <summary>
+    /// Gets or sets an option that specify which components of a party should be included in the result.
+    /// </summary>
+    [JsonPropertyName("includeComponents")]
+    public PartyComponentOptions IncludeComponents { get; set; } = PartyComponentOptions.None;
```
Alxandr, line 33: This almost definitely does not belong in `PartyLookup` - rather, it should probably be a query parameter on the controller-action instead

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PartyName.cs
```diff
@@ -20,8 +20,15 @@ public record PartyName
...
     /// </summary>
     [JsonPropertyName("name")]
     public string? Name { get; set; }
+
+    /// <summary>
+    /// Gets or sets the components of a person's name for this result.
+    /// </summary>
+    [JsonPropertyName("personNameComponents")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public PersonNameComponents? PersonNameComponents { get; set; }
```
Alxandr, line 33: Just calling the property `personName` is probably good enough.

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PersonNameComponents.cs
Alxandr: And how exactly does it do that?

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +400,118 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
     }
+
+    /// <summary>
+    /// Tests the PostPartyNamesLookup with valid input and verifies that the name components are correctly processed.
+    /// </summary>
+    /// <param name="socialSecurityNumber">The Social Security Number.</param>
+    /// <param name="nameComponent">The name component.</param>
+    [Theory]
+    [MemberData(nameof(GetPartyLookupTestData))]
+    public async Task PostPartyNamesLookup_ValidInput_NameComponents_OK(string[] socialSecurityNumber, PartyComponentOptions nameComponent)
```
Alxandr, line 411: Given that this is really testing a HTTP endpoint, rather than having a `PartyComponentOptions` as a input, it would make more sense to take in a query string value. There are 3 different values that should be tested:

* `` (empty string)
* `?includeComponents=` (empty parameter)
* `?includeComponents=person-name`

If you need help with the implementation, the resource registry has a model binder for a similar enum:

* [Model binder](https://github.com/Altinn/altinn-resource-registry/blob/290e0c3af880f5c0a501b196c5fa31498ec220c0/src/Altinn.ResourceRegistry/Models/ModelBinding/AccessListIncludesModelBinder.cs#L14)
* [Swashbuckle schema filter](https://github.com/Altinn/altinn-resource-registry/blob/290e0c3af880f5c0a501b196c5fa31498ec220c0/src/Altinn.ResourceRegistry/Models/ApiDescriptions/AccessListIncludesSchemaFilter.cs#L14)

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
Ahmed-Ghanam: This was code I copied from a similar test unit. I removed it because we do not test caching.

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
Ahmed-Ghanam: I modified the unit test to test the result using valid inputs

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +399,116 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
+    /// Provides valid test data for testing party lookup functionality using different component options.
+    /// </summary>
+    /// <returns>A collection of test data, each containing two social security numbers and corresponding component option.</returns>
+    public static TheoryData<string[], string> GetPartyLookupValidTestData()
+    {
+        return new TheoryData<string[], string>
+        {
+            { ["01039012345","01017512345"], null },
+            { ["01039012345","01017512345"] , string.Empty },
+            { ["01039012345","01017512345"], "?includeComponents=0" },
```
Alxandr, line 510: These are horrible API options. We should not make an API that requires sending arbitrary numbers that have no apparently semantic meaning.

Ahmed-Ghanam: These are test data intended for building queries. If you have an example of well-structured/not-horrible test data, it would be greatly appreciated.

Alxandr: This is a flags enum. You can't just expect the values to be used as strings. I already explained what I expect to work, but to expand on it further:

* (empty string, no query): `PartyComponentOptions.None`
* `?includeComponents=` (that is, the query value is an empty string/null): `PartyComponentOptions.None`
* `?includeComponents=person-name`: `PartyComponentOrder.NameComponents`
* `?includeComponents=none`: bad request
* `?includeComponents=anything-else-really`: bad request

I care much more that the HTTP API is well designed than how it's actually implemented - cause the implementation can always be changed later. But the API is basically locked once it's merged.

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +399,176 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
+
+        int sblEndpointInvoked = 0;
+        HttpRequestMessage sblRequest = null;
+        DelegatingHandlerStub messageHandler = new((request, cancellationToken) =>
+        {
+            sblRequest = request;
+            sblEndpointInvoked++;
+
+            // Simulate an error response for invalid input
+            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.BadRequest));
```
SandGrainOne, line 426: This message handler is for simulating the response from SBLBridge. In this test the the call chain should never get that far. The invalid input is detected long before a call to SBLBridge is being made. 

If we want to verify this properly we should avoid returning BadRequest here as that could confuse the intention of the test. Especially the comment. Could we return something else or maybe throw an exception instead? Remove the comment.

Another options could be to simply just not provide a message handler.

SandGrainOne (commented):
Some findings from me.

src/Altinn.Register/src/Altinn.Register/Clients/PartiesClient.cs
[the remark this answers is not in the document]
Ahmed-Ghanam: We have already fixed this issue. Here’s how it works now:
Data Retrieval: When we need information about a person, we get all their data from our source, including their name components.
Caching: We save (cache) all this data so we don’t have to fetch it again.
On-Demand Name Retrieval: If someone asks for the name parts later, we simply pull them from the cached data.
This way, we always have the name parts ready to return without needing to fetch the data again.

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +399,176 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
+            });
+        });
+
+        await Task.WhenAll(loadTasks);
+
+        PartyNamesLookupResult expectedResult = new()
+        {
+            PartyNames = nameComponentOption switch
+            {
+                "?includeComponents=nameComponents" => testPartyNames.Where(e => socialSecurityNumbers.Contains(e.Ssn))
```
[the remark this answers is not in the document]
Ahmed-Ghanam: @Alxandr recommends utilizing an enum parser to circumvent the direct use of enum member values and names. He provided an example demonstrating how specific text can be bound to enum members. Consequently, we have agreed to use the following query string:
- `?includeComponents=person-name`: Name components will be populated
- `?includeComponents=none`: Bad request
- `?includeComponents=anything-else-else`: bad request

src/Altinn.Register/src/Altinn.Register/Clients/PartiesClient.cs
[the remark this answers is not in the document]
SandGrainOne: We're making the assumption that we will continue to always include name components regadless of input parameter. This will change in the future.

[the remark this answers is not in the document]
Alxandr: This code is just the A2 bridge implementation. It will all be changed once we go against the A3 register db.

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
Ahmed-Ghanam: These use cases have been implemented and discussed in another comment. Closing this one.

src/Altinn.Platform.Models/src/Altinn.Platform.Models/Register/PartyComponentOption.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Altinn.Platform.Register.Models;
+
+/// <summary>
+/// Specifies the components that should be included when retrieving party's information.
+/// </summary>
+[Flags]
+public enum PartyComponentOption : uint
```
Alxandr, line 7: `Flags` enums should be named plurally, `PartyComponentOptions`.

src/Altinn.Register/src/Altinn.Register.Core/Parties/IV1PartyService.cs
```diff
@@ -40,12 +42,13 @@ public interface IV1PartyService
...
     /// <summary>
-    /// Lookup party names by social security number organization number.
+    /// Lookup party names by social security number or organization number.
     /// </summary>
     /// <param name="lookupValues">The set of ssn/org.nr.</param>
+    /// <param name="partyComponentOption">Specifies the components that should be included when retrieving party's information.</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>An async enumerable of <see cref="V1Models.PartyName"/>.</returns>
-    IAsyncEnumerable<V1Models.PartyName> LookupPartyNames(IEnumerable<V1Models.PartyLookup> lookupValues, CancellationToken cancellationToken = default);
+    IAsyncEnumerable<PartyName> LookupPartyNames(IEnumerable<PartyLookup> lookupValues, PartyComponentOption partyComponentOption, CancellationToken cancellationToken = default);
```
Alxandr, line 51: Please keep the `V1Models` - there is a reason it was introduced here as preparation for V2 models.

src/Altinn.Register/src/Altinn.Register/ApiDescriptions/AccessListIncludesSchemaFilter.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.OpenApi.Models;
+
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace Altinn.Register.ApiDescriptions;
+
+/// <summary>
+/// Schema filter for <see cref="PartyComponentOption"/>.
+/// </summary>
+public sealed class AccessListIncludesSchemaFilter
```
Alxandr:
```suggestion
public sealed class PartyComponentOptionSchemaFilter
```

src/Altinn.Register/test/Altinn.Register.Tests/IntegrationTests/PartiesControllerTests.cs
```diff
@@ -398,4 +399,161 @@ public async Task PostPartyNamesLookup_PartialInvalidInput_OK()
...
+
+    /// <summary>
+    /// Provides valid test data for testing party lookup functionality using different component options.
+    /// </summary>
+    /// <returns>A collection of test data, each containing two social security numbers and corresponding component option.</returns>
+    public static TheoryData<string[], string> GetPartyLookupValidTestData()
+    {
+        return new TheoryData<string[], string>
+        {
+            { ["01039012345","01017512345"], null },
```
Alxandr, line 540: You can remove `null` here as it's identical to `string.Empty` since it's only used to build up a query-string.

Alxandr (approved):

PR #332, opened 2024-11-13
feat: add lease records

Alxandr:
Store when a lease was last acquired and released to be able to do scheduling decisions based on it.


TheTechArch (approved):
LGTM

PR #339, opened 2024-11-14
chore: improve handling of shared resources in tests

Alxandr:
Minor improvement to how shared resources are handled in tests. This increases the determinism of cleanup of async resources in tests which was previously just scheduled to run at a later time, while not re-introducing the issue of database-servers not being reused due to tests running in serial.


simen-rekkedal (approved):
Lgtm

PR #357, opened 2024-12-09
feat: wait for bus

Alxandr:


TheTechArch (approved):
LGTM Almost like #ruter app

PR #368, opened 2024-12-19
chore: improve import tracking more

Alxandr:
Adds simple caching to the import-tracker, and uses a channel and background worker to reduce concurrency.


simen-rekkedal (approved):
LGTM

PR #369, opened 2024-12-20
feat: consumer meters

Alxandr:


simen-rekkedal (approved):
LGTM, 100% coverate actually

PR #372, opened 2025-01-14
feat: A2 import flow

Alxandr:


TheTechArch (approved):
Great code. Learned some new concept. Did not understand how the foreach on the batch worked but I assume it is some mass transit magic. (or just lack of understanding from my side)

PR #389, opened 2025-02-04
chore: make all DB calls async

Alxandr:
image


TheTechArch (approved):
var feedback = await LGTM

PR #390, opened 2025-02-04
feat: read external role definitions

Alxandr:


TheTechArch (approved):
Thanks for learning me about ordinal and possible performance improvement of looking up on name. LGTM.  I assume Cache is a .net framework class. Was not able to find the import

PR #410, opened 2025-02-18
feat: upsert external role assignments command

Alxandr:


TheTechArch (approved):
Impressive amount of code as always. On a high level it seems to do the role import functionality through a batch, and you have different utilities to track progress and handle errors. LGTM

PR #440, opened 2025-03-20
chore: party gone

Alxandr:


TheTechArch (approved):
LGTM but will not happen since no changes in Altinn 2

PR #441, opened 2025-03-21
chore: handle resilience different for A2 import service

Alxandr:


TheTechArch (approved):
Interesting how you tested this. LGTM

PR #510, opened 2025-07-08
chore: disable masstransit during init-only

Alxandr:


TheTechArch (approved):
(A jaunty developer-folk tune, capoed at 2nd fret)

Verse 1
Well I opened up my IDE, on a rainy Monday morn
Spotted MassTransit spinning up, and I thought “this can’t be born”
Two recurring jobs were calling, every sixty ticking seconds
But during init, that engine hum––I just couldn’t stand the beckons

Pre-Chorus
There’s a Quirk in Altinn’s heart, a flag that led me astray
When party imports fired themselves, I’d lose half my day

Chorus
So I clicked “Create Pull Request,” with a trembling little grin
“Disable MassTransit on init,” I said, “Let’s save some CPU spin”
Two if checks in the config now, have cast that job aside
Oh, we’ve silenced that recurring hum—our pipelines now can glide!

Verse 2
First was A2 PartyImport, the code that loves to loop
Then A2 PartyUserId, another hop-and-scoop
Both wrapped in GetValue<bool>, toggles shining true or false
But when I flipped them off at dawn, I found my system’s waltz

Pre-Chorus
No more queues in early hours, no more needless bus demands
Just a quiet, peaceful init time, yeah that felt pretty grand

Chorus
So I clicked “Create Pull Request,” with a trembling little grin
“Disable MassTransit on init,” I said, “Let’s save some CPU spin”
Two if checks in the config now, have cast that job aside
Oh, we’ve silenced that recurring hum—our pipelines now can glide!

Bridge
Static lambdas waiting for the bus, ValueTasks in the wings
But until we truly need them, we’ll let our background rest and sing
LeaseName set to A2, but only when the flag’s alive
For early startup peace of mind, we let those toggles drive

Chorus (soft)
I raised a review comment, “Look at Mom—no jobs to start!”
In the diff you’ll see my changes, I disabled all that heart
But come deployment, flip it back, if you need that minute beat
Till then we’ll savor idle time, in our Docker-lit retreat

Outro
So here’s to clever toggles, and pull requests that save the day
Disabling MassTransit on init, so we code in calm array
May your mornings be muted, your logs be ever clean
And your CI pipelines hum along, but not until they’re seen.

PR #538, opened 2025-08-27
chore: log cleanup info

Alxandr:


TheTechArch (approved):
The logger speaks, its tale is clear,
Of queues untouched, with none to fear.
“No pages new,” it softly sighs,
“The run was calm beneath the skies.”

Before the job, the count stood tall,
After the job — unchanged at all.
StartEnqueuedMax and EndEnqueuedMax,
Two sentinels guarding silent tracks.

But then, a whisper stirs the stream,
A call awakens a cleaner’s dream:
"Starting periodic party cleanup!" — loud,
The logs declare it, strong and proud.

No page was queued, no mess to sweep,
Yet still the code its promise keeps.
For in this world of tasks and loops,
Even silence echoes through the groups.

PR #587, opened 2025-10-03
fix: party mapping exception messages

Alxandr:
---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #587 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #586


TheTechArch (approved):
Missing Bachelor but ok.

PR #796, opened 2026-04-20
fix: map username to party.Name for SI users

Alxandr:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated party name resolution for self-identified users to use correct field mapping with appropriate fallback values, ensuring accurate data display in the registration system.


simen-rekkedal (approved):
lgtm

PR #817, opened 2026-04-22
feat: npr import

Alxandr:
## Summary by CodeRabbit

* **New Features**
  * NPR person document parsing/validation including guardianship extraction and richer address handling (international support)
  * Location lookup service with cached country/municipality data
  * Debug endpoints to fetch NPR person data and guardianships for local testing
  * Improved validation feedback for person and address data

* **Chores**
  * New NPR integration project and reorganization of related code
  * Database migrations: country & municipality tables and seed data
  * Expanded test fixtures and test harness for NPR scenarios


src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/ActiveElementValidator.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    validatedList.Add(validatedItem.Value);
+                }
+            }
+
+            if (context.HasErrors)
+            {
+                validated = [];
+                return false;
+            }
```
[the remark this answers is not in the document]
Alxandr: This is how the validation library is designed to be used. Typically, each object gets it's own state though.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/InternationalMailingAddress.cs
```diff
@@ -0,0 +1,75 @@
...
+
+/// <summary>
+/// Represents a structured international mailing address.
+/// </summary>
+public sealed record InternationalMailingAddress
+{
+    /// <summary>
+    /// Gets the care-of name for the address.
+    /// </summary>
+    [JsonPropertyName("coadressenavn")]
```
TheTechArch: Er det feil på skatt sin side at casing er forskjellig sammenlignet med StreetAdress og MartikkelADress

Alxandr: Feil på min side. Vi bruker case-insensitive JSON dog, så det ville fungert tror jeg. Men skal uansett fikse.

TheTechArch:
Full feedback fra claude. Tror mestparten kanskje er feil. 



Review: feat/freg-import — FREG/NPR integration
The branch name says "freg" but the code is named "Npr" (the API is NPR, the data is FREG — same thing in practice). The mapping pipeline is PersonDocument → PersonDocumentValidator → PersonRecord/GuardianshipInfo. Most of the mapping is solid and round-trips cleanly against 15 sample fixtures. The issues below range from real bugs to smells.

Mapping bugs
1. coadressenavn vs coAdressenavn typo in InternationalMailingAddress — probably the most concerning mapping issue.

[InternationalMailingAddress.cs:13](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/InternationalMailingAddress.cs#L13): [JsonPropertyName("coadressenavn")] — lowercase a
[StreetAddress.cs:37](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/StreetAddress.cs#L37) and [MatrikkelAddress.cs:50](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/MatrikkelAddress.cs#L50): "coAdressenavn" — capital A
Real NPR payload in [69914600685/npr.json:124](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/test/Altinn.Register.Tests/Testdata/Npr/Persons/valid/69914600685/npr.json#L124) uses "coAdressenavn".
This may currently work because JsonSerializerDefaults.Web is case-insensitive, but it's inconsistent with the rest of the schema bindings and fragile. Fix the casing.

2. PostgreSqlLocationLookupProvider.Cache gets stuck on first-load failure

In [PostgreSqlLocationLookupProvider.Cache.cs:92-121](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Persistence/Location/PostgreSqlLocationLookupProvider.Cache.cs#L92-L121):


if (!task.IsCompletedSuccessfully)
{
    return task;            // early-returns the faulted task
}

var result = task.GetAwaiter().GetResult();
lock (_lock)
{
    if (ReferenceEquals(_state, current))
    {
        if (task.IsCompletedSuccessfully) { _state = result; ... }
        else { _state = null; }  // DEAD — unreachable after early return
    }
}
If the initial DB load fails (transient outage, timeout), _state keeps pointing at the faulted task forever. Every subsequent caller gets the same error until the process restarts. The else { _state = null; } reset branch is dead code because of the early return above. Move the failure reset before the early return, or drop the early return and let the lock decide.

3. Empty-string postal codes silently become "0000"

[PersonDocumentValidator.cs:1198-1222](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1198-L1222): only null triggers Required. Combined with IsDigitsOnly("") == true ([line 1238-1239](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1238-L1239) — ContainsAnyExcept returns false for empty span), an empty postal code passes validation and gets PadLeft(4, '0') → "0000". Either reject empty in the Required check or in IsDigitsOnly.

4. FreeFormMailingAddress postal-duplication heuristic is brittle

[PersonDocumentValidator.cs:848-852](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L848-L852):


var address = string.Join(' ', input.AddressLines);
if (input.AddressLines.Length > 0 && !input.AddressLines[^1].StartsWith(postalInfo.Code))
{
    address = $"{address} {postalInfo.Code} {postalInfo.Name}".Trim();
}
postalInfo.Code has been padded to 4 digits. If an NPR line is "123 Oslo" and the raw code was "123" (padded to "0123"), StartsWith("0123") returns false and the block duplicates the postal info into the address. StartsWith also has no StringComparison. Comparable issue when the original line contains the postal code in the middle, not at the start.

5. Guardianship detection for c/o-prefix uses the unvalidated raw array

[PersonDocumentValidator.cs:236-240](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L236-L240):


var hasGuardian = !input.GuardianshipOrPowerOfAttorney.IsDefaultOrEmpty;
var prefix = hasGuardian ? "v/" : "c/o ";
IsDefaultOrEmpty is tested on the raw NPR array, which contains both verge and fremtidsfullmakt entries (see [PersonDocument.cs:73-74](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocument.cs#L73-L74) — the field is called GuardianshipOrPowerOfAttorney). A person with only a future power of attorney — no actual verge — will get "v/" prefixed. Either check whether any entry has Guardianship is not null, or pass the validated guardianship list instead.

6. UnknownResidentialAddress drops bostedskommune

[ResidentialAddressElement.cs:41-48](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/ResidentialAddressElement.cs#L41-L48) models ukjentBosted.bostedskommune, but both residential-address validators ([PersonDocumentValidator.cs:667-676](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L667-L676) and [PersonDocumentValidator.cs:729-736](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L729-L736)) return fully-null records. If the municipality is the one piece of signal there, it should land in MunicipalNumber/MunicipalName.

7. City trimming inconsistency between free-form and structured international addresses

Structured [PersonDocumentValidator.cs:1054](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1054): City = input.CityOrPlaceName (untrimmed)
Free-form [PersonDocumentValidator.cs:1000](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1000): City = cityOrPlaceName (trimmed)
Pick one.

8. Matrikkel fallback if (postal.Code == null) is dead code

[PersonDocumentValidator.cs:1086-1089](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1086-L1089): after the if (context.HasErrors) return false; guard at L1077, postal.Code is always set. The "0000" fallback is unreachable. Remove it, or replace the guard with logic that actually permits continuing without a postal area.

Code-quality / bad practice
9. MaskinPortenClient has a -10 seconds JWT iat skew and commented-out debug logging left in

[MaskinPortenClient.cs:216](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/ServiceDefaults.HttpClient.MaskinPorten/MaskinPortenClient.cs#L216):


var now = _timeProvider.GetUtcNow() - TimeSpan.FromSeconds(10);
And [MaskinPortenClient.cs:122-130](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/ServiceDefaults.HttpClient.MaskinPorten/MaskinPortenClient.cs#L122-L130) contains a commented-out #if DEBUG _logger.LogWarning(..., jwtAssertion) block. Both look like local-dev scaffolding that slipped in. If the 10-second backdate is a real clock-skew mitigation, it belongs in a named constant with a comment explaining why; if it isn't, revert it. Either way, remove the commented JWT-leak log.

10. DebugController.Test (npr/person/{fnr}) is [AllowAnonymous] inside an [Authorize(Policy = "Debug")] controller

[DebugController.cs:49-83](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs#L49-L83). It's guarded by [DevTestCondition] so the route should never register outside dev/test, but the combination is a footgun — the endpoint proxies full NPR responses. Either drop [AllowAnonymous] (Test2 at [line 88-126](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs#L88-L126) already requires Debug policy for the same kind of data) or add a comment explaining why anonymous is safe here.

11. GuardianshipRoleMapper.TryFindRoleByNprValues returns true with role == null

[PersonDocumentValidator.cs:170-176](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L170-L176) depends on this subtle three-state API ("unknown pair" = false; "known but intentionally unmapped" = true + null; "mapped" = true + value). This is easy to misuse. An enum return (Unknown | Ignored | Mapped) would make the contract explicit.

12. ImmutableValueSet<T> throws NotImplementedException on most IReadOnlySet operations

[ImmutableValueSet.cs:144-178](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/ImmutableValueSet.cs#L144-L178). There's a // TODO: Move this to modelutils and build out properly — fine for now, but anyone casting to IReadOnlySet<T> and calling SetEquals/Overlaps/etc. will crash. Worth tracking rather than leaving as an unowned TODO.

13. Minor smells

[PersonDocumentValidator.cs:1083](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1083) and [:1308](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1308) use new List<string>() while the rest of the file uses collection expressions []. Trivial.
[Optional<T>](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/Optional.cs) struct doesn't override Equals/GetHashCode. Fine today but will bite if someone ever puts one in a dictionary.
[DebugController.cs:75](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs#L75) hardcodes a ~25-element part=...&part=... query string. Trivial, dev-only.
The "dev: local npr proxy" commit at the base of the branch shouldn't land on main; if you're going to merge it, rebase/squash so the JWT-skew change is owned by a more explanatory commit.
What looks fine
PersonDocument → PersonRecord mapping round-trips for all 15 fixtures; display/short-name rules, municipal-number preservation (string form with leading zeros), municipality-lookup fallback to integer keys, and the matrikkel cadastral-number formatting all check out.
ActiveElement<T> JSON converter correctly detects 0/1/many active entries and surfaces "multiple" as a validation error.
The new v0.54-country-municipality migration matches the lookup shape, and DataSourceMappingsTests.MapsMunicipalityStatus exercises the enum mapping.
GuardianshipRoleMapper relocation to the new Integrations.Npr project is clean; the end: whitespace diff noise in Guardianships.g.cs is just the generator's output format change.
Suggested top-3 fixes before merge
Fix the coadressenavn casing in InternationalMailingAddress.
Make the location-lookup cache actually recover from a failed first load.
Decide whether the Maskinporten -10s skew and commented-out JWT log are intentional, then either document them or revert.Review: feat/freg-import — FREG/NPR integration
The branch name says "freg" but the code is named "Npr" (the API is NPR, the data is FREG — same thing in practice). The mapping pipeline is PersonDocument → PersonDocumentValidator → PersonRecord/GuardianshipInfo. Most of the mapping is solid and round-trips cleanly against 15 sample fixtures. The issues below range from real bugs to smells.

Mapping bugs
1. coadressenavn vs coAdressenavn typo in InternationalMailingAddress — probably the most concerning mapping issue.

[InternationalMailingAddress.cs:13](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/InternationalMailingAddress.cs#L13): [JsonPropertyName("coadressenavn")] — lowercase a
[StreetAddress.cs:37](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/StreetAddress.cs#L37) and [MatrikkelAddress.cs:50](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/MatrikkelAddress.cs#L50): "coAdressenavn" — capital A
Real NPR payload in [69914600685/npr.json:124](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/test/Altinn.Register.Tests/Testdata/Npr/Persons/valid/69914600685/npr.json#L124) uses "coAdressenavn".
This may currently work because JsonSerializerDefaults.Web is case-insensitive, but it's inconsistent with the rest of the schema bindings and fragile. Fix the casing.

2. PostgreSqlLocationLookupProvider.Cache gets stuck on first-load failure

In [PostgreSqlLocationLookupProvider.Cache.cs:92-121](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Persistence/Location/PostgreSqlLocationLookupProvider.Cache.cs#L92-L121):


if (!task.IsCompletedSuccessfully)
{
    return task;            // early-returns the faulted task
}

var result = task.GetAwaiter().GetResult();
lock (_lock)
{
    if (ReferenceEquals(_state, current))
    {
        if (task.IsCompletedSuccessfully) { _state = result; ... }
        else { _state = null; }  // DEAD — unreachable after early return
    }
}
If the initial DB load fails (transient outage, timeout), _state keeps pointing at the faulted task forever. Every subsequent caller gets the same error until the process restarts. The else { _state = null; } reset branch is dead code because of the early return above. Move the failure reset before the early return, or drop the early return and let the lock decide.

3. Empty-string postal codes silently become "0000"

[PersonDocumentValidator.cs:1198-1222](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1198-L1222): only null triggers Required. Combined with IsDigitsOnly("") == true ([line 1238-1239](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1238-L1239) — ContainsAnyExcept returns false for empty span), an empty postal code passes validation and gets PadLeft(4, '0') → "0000". Either reject empty in the Required check or in IsDigitsOnly.

4. FreeFormMailingAddress postal-duplication heuristic is brittle

[PersonDocumentValidator.cs:848-852](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L848-L852):


var address = string.Join(' ', input.AddressLines);
if (input.AddressLines.Length > 0 && !input.AddressLines[^1].StartsWith(postalInfo.Code))
{
    address = $"{address} {postalInfo.Code} {postalInfo.Name}".Trim();
}
postalInfo.Code has been padded to 4 digits. If an NPR line is "123 Oslo" and the raw code was "123" (padded to "0123"), StartsWith("0123") returns false and the block duplicates the postal info into the address. StartsWith also has no StringComparison. Comparable issue when the original line contains the postal code in the middle, not at the start.

5. Guardianship detection for c/o-prefix uses the unvalidated raw array

[PersonDocumentValidator.cs:236-240](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L236-L240):


var hasGuardian = !input.GuardianshipOrPowerOfAttorney.IsDefaultOrEmpty;
var prefix = hasGuardian ? "v/" : "c/o ";
IsDefaultOrEmpty is tested on the raw NPR array, which contains both verge and fremtidsfullmakt entries (see [PersonDocument.cs:73-74](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocument.cs#L73-L74) — the field is called GuardianshipOrPowerOfAttorney). A person with only a future power of attorney — no actual verge — will get "v/" prefixed. Either check whether any entry has Guardianship is not null, or pass the validated guardianship list instead.

6. UnknownResidentialAddress drops bostedskommune

[ResidentialAddressElement.cs:41-48](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/ResidentialAddressElement.cs#L41-L48) models ukjentBosted.bostedskommune, but both residential-address validators ([PersonDocumentValidator.cs:667-676](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L667-L676) and [PersonDocumentValidator.cs:729-736](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L729-L736)) return fully-null records. If the municipality is the one piece of signal there, it should land in MunicipalNumber/MunicipalName.

7. City trimming inconsistency between free-form and structured international addresses

Structured [PersonDocumentValidator.cs:1054](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1054): City = input.CityOrPlaceName (untrimmed)
Free-form [PersonDocumentValidator.cs:1000](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1000): City = cityOrPlaceName (trimmed)
Pick one.

8. Matrikkel fallback if (postal.Code == null) is dead code

[PersonDocumentValidator.cs:1086-1089](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1086-L1089): after the if (context.HasErrors) return false; guard at L1077, postal.Code is always set. The "0000" fallback is unreachable. Remove it, or replace the guard with logic that actually permits continuing without a postal area.

Code-quality / bad practice
9. MaskinPortenClient has a -10 seconds JWT iat skew and commented-out debug logging left in

[MaskinPortenClient.cs:216](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/ServiceDefaults.HttpClient.MaskinPorten/MaskinPortenClient.cs#L216):


var now = _timeProvider.GetUtcNow() - TimeSpan.FromSeconds(10);
And [MaskinPortenClient.cs:122-130](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/ServiceDefaults.HttpClient.MaskinPorten/MaskinPortenClient.cs#L122-L130) contains a commented-out #if DEBUG _logger.LogWarning(..., jwtAssertion) block. Both look like local-dev scaffolding that slipped in. If the 10-second backdate is a real clock-skew mitigation, it belongs in a named constant with a comment explaining why; if it isn't, revert it. Either way, remove the commented JWT-leak log.

10. DebugController.Test (npr/person/{fnr}) is [AllowAnonymous] inside an [Authorize(Policy = "Debug")] controller

[DebugController.cs:49-83](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs#L49-L83). It's guarded by [DevTestCondition] so the route should never register outside dev/test, but the combination is a footgun — the endpoint proxies full NPR responses. Either drop [AllowAnonymous] (Test2 at [line 88-126](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs#L88-L126) already requires Debug policy for the same kind of data) or add a comment explaining why anonymous is safe here.

11. GuardianshipRoleMapper.TryFindRoleByNprValues returns true with role == null

[PersonDocumentValidator.cs:170-176](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L170-L176) depends on this subtle three-state API ("unknown pair" = false; "known but intentionally unmapped" = true + null; "mapped" = true + value). This is easy to misuse. An enum return (Unknown | Ignored | Mapped) would make the contract explicit.

12. ImmutableValueSet<T> throws NotImplementedException on most IReadOnlySet operations

[ImmutableValueSet.cs:144-178](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/ImmutableValueSet.cs#L144-L178). There's a // TODO: Move this to modelutils and build out properly — fine for now, but anyone casting to IReadOnlySet<T> and calling SetEquals/Overlaps/etc. will crash. Worth tracking rather than leaving as an unowned TODO.

13. Minor smells

[PersonDocumentValidator.cs:1083](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1083) and [:1308](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/PersonDocumentValidator.cs#L1308) use new List<string>() while the rest of the file uses collection expressions []. Trivial.
[Optional<T>](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register.Integrations.Npr/Person/Optional.cs) struct doesn't override Equals/GetHashCode. Fine today but will bite if someone ever puts one in a dictionary.
[DebugController.cs:75](vscode-webview://1p7u1qvoqfs24b1cui0s97613epab1jqnc767564q0rsehesalen/src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs#L75) hardcodes a ~25-element part=...&part=... query string. Trivial, dev-only.
The "dev: local npr proxy" commit at the base of the branch shouldn't land on main; if you're going to merge it, rebase/squash so the JWT-skew change is owned by a more explanatory commit.
What looks fine
PersonDocument → PersonRecord mapping round-trips for all 15 fixtures; display/short-name rules, municipal-number preservation (string form with leading zeros), municipality-lookup fallback to integer keys, and the matrikkel cadastral-number formatting all check out.
ActiveElement<T> JSON converter correctly detects 0/1/many active entries and surfaces "multiple" as a validation error.
The new v0.54-country-municipality migration matches the lookup shape, and DataSourceMappingsTests.MapsMunicipalityStatus exercises the enum mapping.
GuardianshipRoleMapper relocation to the new Integrations.Npr project is clean; the end: whitespace diff noise in Guardianships.g.cs is just the generator's output format change.
Suggested top-3 fixes before merge
Fix the coadressenavn casing in InternationalMailingAddress.
Make the location-lookup cache actually recover from a failed first load.
Decide whether the Maskinporten -10s skew and commented-out JWT log are intentional, then either document them or revert.

TheTechArch (approved):

TheTechArch:
Updated from claude

## Resolved since previous review

- Conflicts: gone, working tree clean.
- `internal sealed class NprClient`: `partial` dropped.
- JSON deserialization is wrapped in `try/catch (JsonException)`: malformed payloads now return `Problems.PartyFetchFailed` with the JSON error in extensions instead of throwing.
- `ImmutableValueSet.UnionWith` doc now correctly states the result is sorted by the set's comparer.
- Typo fixed: `Impementation` → `Implementation`.
- Better `.csproj` choice: `Microsoft.Extensions.Http` package replaces the `FrameworkReference Microsoft.AspNetCore.App` I flagged. Lighter dependency — good call. Also added to `Directory.Packages.props`.
- New `accept-npr-valid-results` is just a recipe for promoting `.received.json` → `.validated.json` after regenerating golden files. Nice ergonomics.

## Overview

Refactor of the NPR integration: hand-rolled JSON parsing is gone, the client now fetches a full person document and returns a single `NprPerson`, with `INprClient` exposed from Core.

- Removes `PartyImport/Npr/NprClient.cs` — 370 lines of bespoke `Utf8JsonReader` and two custom converters — plus its 201-line unit-test file.
- Adds `INprClient` and a thin `NprClient` that defers to `PersonDocumentValidator`.
- New domain types `NprPerson` / `NprGuardianship` replace the two-headed `PersonRecord + ImmutableArray<GuardianshipInfo>` return shapes.
- `ImmutableValueSet<T>` moves from `Integrations.Npr` to `Core/Utils` with a new `UnionWith`.
- Test data: 14 person cases collapse `party.validated.json` + `guardianships.validated.json` into a single `result.validated.json`, matching the unified return type.

## Strengths

- Removing the hand-rolled `JsonConverters` is a big win — the existing `PersonDocumentValidator` already handles structural validation correctly, so this is pure subtraction.
- `NprPerson` no longer pretends to know Altinn identifiers. No more `FieldValue.Unset` placeholders for `PartyUuid`, `PartyId`, `User`, `ExternalUrn`, `Source`, `CreatedAt`, `ModifiedAt`, or `VersionId`. The merging into `PersonRecord` happens in `NprEnricher`, where it belongs.
- `INprClient` lives in Core, so consumers (`NprEnricher`, `DebugController`) and the test fakes (`FakeHttpHandlers.For<INprClient>()`) depend on the abstraction, not the integration project.
- DI extension method `NprIntegrationServiceCollectionExtensions` gives host code a clean `services.AddNprClient()` entry point.
- The `try/catch (JsonException)` on the response body is the right move for an external API.

## Behavior changes worth flagging

- `NprEnricher` now overwrites name/address/birth/death on the existing `PersonRecord` (`NprEnricher.cs:64-76`). Previously the enricher only added guardianships. NPR is now authoritative for:
  - `DisplayName`
  - `FirstName`
  - `MiddleName`
  - `LastName`
  - `ShortName`
  - `Address`
  - `MailingAddress`
  - `DateOfBirth`
  - `DateOfDeath`

  Presumably this is the intent of the PR, but it is not in the description and is the highest-risk change here. Worth a sentence in the PR body so reviewers and ops know.

- `404` maps to `Problems.PersonNotFound` instead of `Problems.ReferencedPartyNotFound`. The enricher's "skip when not in NPR" branch was updated; if any other consumer pattern-matched on the old code, it will silently stop firing.

- `PersonIdentifier` reassignment in the `with` expression at `NprEnricher.cs:69` is redundant — the equality check four lines above guarantees it is the same value. Cosmetic.

## Issues / suggestions

### Simplify `MergeGuardianshipsForSameGuardian`

`MergeGuardianshipsForSameGuardian` does too much work (`PersonDocumentValidator.cs:1440-1481`).

The builder is already sorted by `SortGuardianships` before this is called, so duplicates are adjacent. Current implementation does a `HashSet` pre-scan, then `GroupBy`, then re-sorts.

A single linear pass coalescing adjacent equal guardians would be simpler and faster, and the trailing `SortGuardianships` call could go.

### Avoid allocations in `SortGuardianships`

`SortGuardianships` allocates two strings per comparison:

```csharp
string.CompareOrdinal(g1.Guardian.ToString(), g2.Guardian.ToString())

If PersonIdentifier is comparable, or its underlying Value is, compare directly to avoid the per-comparison allocations.

Restore direct HTTP/error-path coverage for NprClient.GetPerson

There are no direct tests for the new client's HTTP/error paths.

Deleting NprClientTests.cs removed coverage for:

404 → PersonNotFound
5xx → PartyFetchFailed
null body → PartyFetchFailed
JsonException → PartyFetchFailed

PartyImportFlowTests exercises the 404 path indirectly; the others are uncovered. A small focused test file for NprClient.GetPerson would protect the contract going forward.

Avoid relying on typed-client name in DebugController

DebugController raw endpoint is fragile (DebugController.cs:73).

It does:

httpClientFactory.CreateClient(nameof(INprClient))

That only works because AddHttpClient<INprClient, NprClient>() happens to register a named client under that exact name. If the registration ever changes, the debug endpoint silently breaks or returns a misconfigured HttpClient.

Either inject INprClient and add a GetRawResponseAsync method, or pin the client name to a documented constant.

Revisit stale TODO in ImmutableValueSet

There is a stale comment:

// TODO: Move this to modelutils

Now that the type lives in Core/Utils and has a real consumer, decide whether modelutils is still the destination.

If yes, leave the TODO. If no, drop it.

Implement or avoid incomplete IReadOnlySet<T> members

ImmutableValueSet<T>.IReadOnlySet<T> members all throw new NotImplementedException.

Treating a public type as IReadOnlySet<T> will blow up on:

IsSubsetOf
Overlaps
SetEquals
and related members

Callers today only use Contains, enumeration, and UnionWith, so it is fine in practice. But anyone who passes one of these to a generic API expecting IReadOnlySet<T> will get a runtime crash.

Consider implementing the members, making the type only conditionally implement the interface, or at least leaving a clear // TODO: implement IReadOnlySet<T> members.

Tests
Golden-file approach in NprPersonParsingTests is consolidated cleanly.
The JSON encoder change to JavaScriptEncoder.Create(BasicLatin, Latin1Supplement) keeps Norwegian characters diff-friendly.
The new accept-npr-valid-results recipe makes regenerating goldens a one-liner.
PartyImportFlowTests correctly extends the fake NPR response with status, foedsel, and navn parts now that the enricher pulls the full document.
Sonar previously reported quality gate passed at 73% new-code coverage; would be good to confirm that has not regressed after the conflict resolution.
Summary

Solid direction, and the new commit fixes essentially everything I flagged on the integration boundary:

sealed class cleanup
JSON exception handling
doc accuracy
lighter .csproj dependency

What is left is mostly polish:

tighten MergeGuardianshipsForSameGuardian
drop the redundant PersonIdentifier re-assignment
add minimal HTTP-path coverage for NprClient.GetPerson
stop relying on the typed-client name in DebugController

The behavior expansion in NprEnricher — NPR now authoritative for name/address/birth/death — deserves a sentence in the PR description before merge.

PR #820, opened 2026-04-24
feat: npr client v2

Alxandr:
## Summary by CodeRabbit

* **New Features**
  * Added NPR client interface for standardized person data fetching.
  * Consolidated person and guardian data into unified records.

* **Refactor**
  * Refactored NPR integration from concrete implementations to interface-based architecture.
  * Restructured test data to reflect consolidated person records with embedded guardians.
  * Updated dependency injection to use extension methods for NPR client registration.

* **Tests**
  * Reorganized test fixtures to align with consolidated data structures.


TheTechArch:
## Resolved since previous review

- Conflicts: gone, working tree clean.
- `internal sealed class NprClient`: `partial` dropped.
- JSON deserialization is wrapped in `try/catch (JsonException)`: malformed payloads now return `Problems.PartyFetchFailed` with the JSON error in extensions instead of throwing.
- `ImmutableValueSet.UnionWith` doc now correctly states the result is sorted by the set's comparer.
- Typo fixed: `Impementation` → `Implementation`.
- Better `.csproj` choice: `Microsoft.Extensions.Http` package replaces the `FrameworkReference Microsoft.AspNetCore.App` I flagged. Lighter dependency — good call. Also added to `Directory.Packages.props`.
- New `accept-npr-valid-results` is just a recipe for promoting `.received.json` → `.validated.json` after regenerating golden files. Nice ergonomics.

## Overview

Refactor of the NPR integration: hand-rolled JSON parsing is gone, the client now fetches a full person document and returns a single `NprPerson`, with `INprClient` exposed from Core.

- Removes `PartyImport/Npr/NprClient.cs` — 370 lines of bespoke `Utf8JsonReader` and two custom converters — plus its 201-line unit-test file.
- Adds `INprClient` and a thin `NprClient` that defers to `PersonDocumentValidator`.
- New domain types `NprPerson` / `NprGuardianship` replace the two-headed `PersonRecord + ImmutableArray<GuardianshipInfo>` return shapes.
- `ImmutableValueSet<T>` moves from `Integrations.Npr` to `Core/Utils` with a new `UnionWith`.
- Test data: 14 person cases collapse `party.validated.json` + `guardianships.validated.json` into a single `result.validated.json`, matching the unified return type.

## Strengths

- Removing the hand-rolled `JsonConverters` is a big win — the existing `PersonDocumentValidator` already handles structural validation correctly, so this is pure subtraction.
- `NprPerson` no longer pretends to know Altinn identifiers. No more `FieldValue.Unset` placeholders for `PartyUuid`, `PartyId`, `User`, `ExternalUrn`, `Source`, `CreatedAt`, `ModifiedAt`, or `VersionId`. The merging into `PersonRecord` happens in `NprEnricher`, where it belongs.
- `INprClient` lives in Core, so consumers (`NprEnricher`, `DebugController`) and the test fakes (`FakeHttpHandlers.For<INprClient>()`) depend on the abstraction, not the integration project.
- DI extension method `NprIntegrationServiceCollectionExtensions` gives host code a clean `services.AddNprClient()` entry point.
- The `try/catch (JsonException)` on the response body is the right move for an external API.

## Behavior changes worth flagging

- `NprEnricher` now overwrites name/address/birth/death on the existing `PersonRecord` (`NprEnricher.cs:64-76`). Previously the enricher only added guardianships. NPR is now authoritative for:
  - `DisplayName`
  - `FirstName`
  - `MiddleName`
  - `LastName`
  - `ShortName`
  - `Address`
  - `MailingAddress`
  - `DateOfBirth`
  - `DateOfDeath`

  Presumably this is the intent of the PR, but it is not in the description and is the highest-risk change here. Worth a sentence in the PR body so reviewers and ops know.

- `404` maps to `Problems.PersonNotFound` instead of `Problems.ReferencedPartyNotFound`. The enricher's "skip when not in NPR" branch was updated; if any other consumer pattern-matched on the old code, it will silently stop firing.

- `PersonIdentifier` reassignment in the `with` expression at `NprEnricher.cs:69` is redundant — the equality check four lines above guarantees it is the same value. Cosmetic.

## Issues / suggestions

### Simplify `MergeGuardianshipsForSameGuardian`

`MergeGuardianshipsForSameGuardian` does too much work (`PersonDocumentValidator.cs:1440-1481`).

The builder is already sorted by `SortGuardianships` before this is called, so duplicates are adjacent. Current implementation does a `HashSet` pre-scan, then `GroupBy`, then re-sorts.

A single linear pass coalescing adjacent equal guardians would be simpler and faster, and the trailing `SortGuardianships` call could go.

### Avoid allocations in `SortGuardianships`

`SortGuardianships` allocates two strings per comparison:

```csharp
string.CompareOrdinal(g1.Guardian.ToString(), g2.Guardian.ToString())

If PersonIdentifier is comparable, or its underlying Value is, compare directly to avoid the per-comparison allocations.

Restore direct HTTP/error-path coverage for NprClient.GetPerson

There are no direct tests for the new client's HTTP/error paths.

Deleting NprClientTests.cs removed coverage for:

404 → PersonNotFound
5xx → PartyFetchFailed
null body → PartyFetchFailed
JsonException → PartyFetchFailed

PartyImportFlowTests exercises the 404 path indirectly; the others are uncovered. A small focused test file for NprClient.GetPerson would protect the contract going forward.

Avoid relying on typed-client name in DebugController

DebugController raw endpoint is fragile (DebugController.cs:73).

It does:

httpClientFactory.CreateClient(nameof(INprClient))

That only works because AddHttpClient<INprClient, NprClient>() happens to register a named client under that exact name. If the registration ever changes, the debug endpoint silently breaks or returns a misconfigured HttpClient.

Either inject INprClient and add a GetRawResponseAsync method, or pin the client name to a documented constant.

Revisit stale TODO in ImmutableValueSet

There is a stale comment:

// TODO: Move this to modelutils

Now that the type lives in Core/Utils and has a real consumer, decide whether modelutils is still the destination.

If yes, leave the TODO. If no, drop it.

Implement or avoid incomplete IReadOnlySet<T> members

ImmutableValueSet<T>.IReadOnlySet<T> members all throw new NotImplementedException.

Treating a public type as IReadOnlySet<T> will blow up on:

IsSubsetOf
Overlaps
SetEquals
and related members

Callers today only use Contains, enumeration, and UnionWith, so it is fine in practice. But anyone who passes one of these to a generic API expecting IReadOnlySet<T> will get a runtime crash.

Consider implementing the members, making the type only conditionally implement the interface, or at least leaving a clear // TODO: implement IReadOnlySet<T> members.

Tests
Golden-file approach in NprPersonParsingTests is consolidated cleanly.
The JSON encoder change to JavaScriptEncoder.Create(BasicLatin, Latin1Supplement) keeps Norwegian characters diff-friendly.
The new accept-npr-valid-results recipe makes regenerating goldens a one-liner.
PartyImportFlowTests correctly extends the fake NPR response with status, foedsel, and navn parts now that the enricher pulls the full document.
Sonar previously reported quality gate passed at 73% new-code coverage; would be good to confirm that has not regressed after the conflict resolution.
Summary

Solid direction, and the new commit fixes essentially everything I flagged on the integration boundary:

sealed class cleanup
JSON exception handling
doc accuracy
lighter .csproj dependency

What is left is mostly polish:

tighten MergeGuardianshipsForSameGuardian
drop the redundant PersonIdentifier re-assignment
add minimal HTTP-path coverage for NprClient.GetPerson
stop relying on the typed-client name in DebugController

The behavior expansion in NprEnricher — NPR now authoritative for name/address/birth/death — deserves a sentence in the PR description before merge.

TheTechArch (approved):
🥇

PR #874, opened 2026-05-15
feat: sire integration

acn-dgopa:
## Related Issue(s)
- #{issue number}

## Verification
- [ ] All tests run green

## Documentation

## Summary by CodeRabbit

* **New Features**
  * SIRE (Skatteetaten) integration added: lookup client, organization model, address normalization, business-relationship mapping and enrichment of organization records.
  * New debug endpoints to fetch raw SIRE payloads, events and trigger per-organization enrichment.

* **Database / Migration**
  * Renamed organization/source enum value from "sdf" to "srs" and added matching enum entries.

* **Tests**
  * Added unit and golden-file parsing tests covering SIRE mapping, form/role mappers and enrichment.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Altinn/altinn-register/pull/874?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/apps/Altinn.Register/src/Altinn.Register.AppHost/Program.cs
```diff
@@ -34,6 +34,8 @@
 // dependencies
 var authentication = builder.AddExternalService("altinn-authentication", $"https://platform.{authenticationEnvFragment}.altinn.cloud/");
 var freg = builder.AddExternalService("folkeregisteret", "https://folkeregisteret-api-konsument.sits.no/");
+var sire = builder.AddExternalService("sire", "http://[::1]:4028/"); ////"https://skatteetatenregistrertselskap.api.skatteetaten-test.no/");
```
Alxandr: This is the wrong address.

src/apps/Altinn.Register/src/Altinn.Register.Core/Parties/OrganizationSource.cs
```diff
@@ -21,4 +21,10 @@ public enum OrganizationSource
     /// </summary>
     [JsonStringEnumMemberName("sdf")]
     BusinessAssessedPartnerships,
+
+    /// <summary>
+    /// Skatteetaten Registered Companies (SIRE).
+    /// </summary>
+    [JsonStringEnumMemberName("srs")]
```
Alxandr: sdf = sire. We don't need a new enum case - but it might make sense to rename the one we have. At least that's been my understanding?

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/ISireClient.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Gets the organization information from SIRE based on the provided <see cref="OrganizationIdentifier"/>.
+    /// </summary>
+    /// <param name="organizationIdentifier">The identifier of the person.</param>
+    /// <param name="cancellationToken">A token to cancel the operation.</param>
+    /// <returns>A <see cref="Result{T}"/> containing the organization information.</returns>
+    Task<Result<SireOrganization>> GetOrganization(OrganizationIdentifier organizationIdentifier, CancellationToken cancellationToken);
+}
+
+///// <summary>
```
Alxandr: Remove commented out code

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireBusinessRelationship.cs
```diff
@@ -0,0 +1,34 @@
...
+
+/// <summary>
+/// Represents a business relationship (virksomhetsrelasjon) as returned by the SIRE API.
+/// </summary>
+public sealed record SireBusinessRelationship
+{
+    /// <summary>
+    /// Gets the relationship type (e.g. "styretsLeder").
+    /// </summary>
+    public required string RelationshipType { get; init; }
```
Alxandr: This does not belong in Core. By the time we get to core, the "relationship type" should have been mapped to a role identifier ("daglig-leder" for instance).

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireOrganization.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <summary>
+    /// Gets whether the organization is deleted.
+    /// </summary>
+    public required bool IsDeleted { get; init; }
+
+    /// <summary>
+    /// Gets the tax liability type (e.g. "selskap", "personligSkattepliktig").
+    /// Used for filtering — personally taxable entities are skipped.
+    /// </summary>
+    public required string? TaxLiabilityType { get; init; }
```
Alxandr: If this is used for filtering it does not belong in core.

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireOrganization.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// <summary>
+    /// Gets the last updated timestamp from the postal address.
+    /// </summary>
+    public required DateTimeOffset? LastUpdated { get; init; }
+
+    /// <summary>
+    /// Gets the business relationships (virksomhetsrelasjon).
+    /// </summary>
+    public required IReadOnlyList<SireBusinessRelationship> BusinessRelationships { get; init; }
```
Alxandr: `IReadOnlyList` is not (necesarily) comparable. That makes equality of the `SireOrganization` record not work.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/BusinessRelationship.cs
```diff
@@ -0,0 +1,40 @@
+using System.Text.Json.Serialization;
+
+namespace Altinn.Register.Integrations.Sire.Organization
```
Alxandr: Use file-scoped namespaces for new files. There are multiple of these, but will leave it to this one comment for now.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,338 @@
...
+    public bool TryValidate(
+        ref ValidationContext context,
+        OrganizationDocument input,
+        [NotNullWhen(true)] out SireOrganization? validated)
+    {
+        if (string.IsNullOrWhiteSpace(input.Identifier))
+        {
+            context.AddChildProblem(StdValidationErrors.Required, "/identifikator");
+            validated = null;
+            return false;
```
Alxandr: Don't generally return on the first error. Instead, validate the other properties, such that errors contains all (or at least most) errors.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,338 @@
...
+                var postBoxPrefix = isForeign ? "Postbox" : "Postboks";
+                return $"{postBoxPrefix} {remainder}".Trim();
+            }
+        }
+
+        return input;
+    }
+
+    private static IReadOnlyList<SireBusinessRelationship> MapBusinessRelationships(
+    IReadOnlyList<BusinessRelationship>? relationships)
```
Alxandr: Indentation error

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,338 @@
...
+                    {
+                        personId = parsedPerson;
+                    }
+                    else if (OrganizationIdentifier.TryParse(relatedId.Value, provider: null, out var parsedOrg))
+                    {
+                        orgId = parsedOrg;
+                    }
+                }
+
+                // Foreign identifiers — not supported in 1st iterasjon, skip
```
Alxandr: Do not skip invalid data. Return errors. We should never accept data we do not understand.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,338 @@
...
+            // RelatedIdentifier is the integration project's model
+            RelatedIdentifier? relatedId = rel.RelatedIdentifier;
+            if (relatedId is { Value: not null })
+            {
+                var isNorwegian = string.Equals(relatedId.CountryCode, "NO", StringComparison.OrdinalIgnoreCase)
+                    || string.IsNullOrEmpty(relatedId.CountryCode);
+
+                if (isNorwegian)
+                {
+                    if (PersonIdentifier.TryParse(relatedId.Value, provider: null, out var parsedPerson))
```
Alxandr: This needs to be changed. Sire includes a field taht dictates what kind of identifier we've received, no? The logic should be:

1. if we've received an org; then try to parse the org. If it's invalid, that's an error.
2. if we've received an pers; then try to parse the pers. If it's invalid, that's an error.
3. Anything else is invalid.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/SireClient.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            document = await response.Content.ReadFromJsonAsync<OrganizationDocument>(_options, cancellationToken);
+        }
+        catch (JsonException ex)
+        {
+            return Problems.PartyFetchFailed.Create(
+                detail: "Response deserialization failed",
+                [
+                    new("organization.source", "sire"),
+                    new("json.error", ex.Message),
```
Alxandr: Could this leak/log person identifiers? And if you've copied this from somewhere, that might be an issue where it's copied from as well.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/SireClient.cs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /// <inheritdoc/>
+    public async Task<Result<SireOrganization>> GetOrganization(OrganizationIdentifier organizationIdentifier, CancellationToken cancellationToken)
+    {
+        using var response = await _client.GetAsync($"v1/digdir/{organizationIdentifier}", cancellationToken);
+
+        if (response.StatusCode == HttpStatusCode.NotFound)
+        {
+            return Problems.OrganizationNotFound.Create([new("organization.source", "sire")]);
```
Alxandr: You can add `organization.id` to all of the errors returned here. Organization identifiers are ok to log, and it will make figuring out issues in app-insights much easier.

src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs
```diff
@@ -117,6 +124,115 @@ public async Task<IActionResult> Test2(
...
+    public async Task<IActionResult> GetSireEvents(
+        [FromServices] IHttpClientFactory httpClientFactory,
+        [FromQuery(Name = "from")] uint sequenceNumber = 0,
+        CancellationToken cancellationToken = default)
+    {
+        using var client = httpClientFactory.CreateClient(nameof(ISireClient));
+        if (sequenceNumber == 0)
+        {
+            var startResponse = await client.GetFromJsonAsync<SireStartModel>("https://sire-events/v1/hendelser/start", cancellationToken);
+            sequenceNumber = startResponse!.SequenceNumber;
```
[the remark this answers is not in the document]
Alxandr: Debug endpoint only available locally to developers. Doesn't matter.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+        Debug.Assert(context.Party is { OrganizationIdentifier.HasValue: true });
+
+        var result = await _sireClient.GetOrganization(context.Party.OrganizationIdentifier.Value, cancellationToken);
+
+        if (result.IsProblem && result.Problem.ErrorCode == Problems.OrganizationNotFound.ErrorCode)
+        {
+            // Organization not found in SIRE — mark as deleted
+            context.Party = ((OrganizationRecord)context.Party) with
+            {
+                UnitStatus = "slettet",
```
Alxandr: I don't think this is correct. I believe "S" is the correct value for deleted orgs. Also, we should really make that an enum at some point...

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+            return;
+        }
+
+        result.EnsureSuccess();
+        var organization = result.Value;
+
+        // Filter personally taxable entities - but i dont see this information on the SIRE organization model, so maybe this is not needed?
+        if (string.Equals(organization.TaxLiabilityType, "personligSkattepliktig", StringComparison.OrdinalIgnoreCase))
+        {
+            return;
```
Alxandr: Just returning here is almost certainly not correct, as that will cuase the (very partial) org to be inserted (and probably fail).

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+    {
+        var relationships = organization.BusinessRelationships;
+
+        if (relationships is not { Count: > 0 })
+        {
+            context.RoleAssignments[ExternalRoleSource.RegisteredWithSkatteetaten] = [];
+            return;
+        }
+
+        // Collect identifiers to resolve → ToPartyUuid
```
Alxandr: You don't need to do this anymore. `context.RoleAssignments` can now accept person-identifiers and org-identifiers directly.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+            .ToDictionary(p => p.PersonIdentifier.Value!, p => p.PartyUuid.Value);
+
+        var orgPartyLookup = relatedParties
+            .Where(p => p.OrganizationIdentifier.HasValue)
+            .ToDictionary(p => p.OrganizationIdentifier.Value, p => p.PartyUuid.Value);
+
+        var mapped = new List<UpsertExternalRoleAssignmentsCommand.Assignment>(relationships.Count);
+        foreach (var rel in relationships)
+        {
+            var roleDef = await _roleDefinitions.TryGetRoleDefinition(
```
Alxandr: This is likely wrong usage of this interface/method. The string argument here is supposed to be an Altinn role identifier.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+                ExternalRoleSource.RegisteredWithSkatteetaten,
+                rel.RelationshipType,
+                cancellationToken);
+
+            if (roleDef is null)
+            {
+                Log.RoleDefinitionNotFound(_logger, rel.RelationshipType, context.PartyUuid);
+                continue;
```
Alxandr: Do not ignore unrecognized data.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            Guid toPartyUuid;
+            if (rel.RelatedPersonIdentifier is { } personId)
+            {
+                if (!personPartyLookup.TryGetValue(personId, out toPartyUuid))
+                {
+                    Log.RelatedPartyNotFound(_logger, rel.RelationshipType, context.PartyUuid, personId.ToString());
+                    continue;
```
Alxandr: Do not ignore unrecognized data.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+                }
+            }
+            else if (rel.RelatedOrganizationIdentifier is { } orgId)
+            {
+                if (!orgPartyLookup.TryGetValue(orgId, out toPartyUuid))
+                {
+                    Log.RelatedPartyNotFound(_logger, rel.RelationshipType, context.PartyUuid, orgId.ToString());
+                    continue;
```
Alxandr: Do not ignore unrecognized data.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,179 @@
...
+            else
+            {
+                continue;
+            }
+
+            mapped.Add(new()
+            {
+                Identifier = roleDef.Identifier,
+                ToPartyUuid = toPartyUuid,    // ← the related party
+                ////FromPartyUuid is context.PartyUuid, set by the saga when publishing UpsertExternalRoleAssignmentsCommand
```
Alxandr: This comment can be removed.

src/apps/Altinn.Register/src/Altinn.Register/RegisterHost.cs
```diff
@@ -119,7 +121,7 @@ public static WebApplication Create(string[] args)
             })
             .ValidateDataAnnotations();
 
-        services.AddSingleton<IAuthorizationHandler, AccessTokenHandler>();
+        ////services.AddSingleton<IAuthorizationHandler, AccessTokenHandler>();
```
Alxandr: This should definitely not be committed.

src/apps/Altinn.Register/src/Altinn.Register/RegisterHost.cs
```diff
@@ -189,6 +189,14 @@ public static WebApplication Create(string[] args)
             .ConfigureBaseAddress("https://folkeregisteret/")
             .AddMaskinPortenHandler("register-freg");
 
+        services.AddHttpClient<ISireClient, SireClient>()
+            .ConfigureBaseAddress("http://sire/")
```
Alxandr: Sire should be https.

src/apps/Altinn.Register/src/Altinn.Register/RegisterHost.cs
```diff
@@ -189,6 +189,14 @@ public static WebApplication Create(string[] args)
...
+            .ConfigureHttpClient(client =>
```
Alxandr: This was only for local testing and needs to be removed.

src/apps/Altinn.Register/src/Altinn.Register.Persistence/Migration/v0.55-add-organization-source-srs/00-org-source-enum.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE register.organization_source ADD VALUE IF NOT EXISTS 'srs';
```
Alxandr: I'm pretty sure we don't need a new enum value. We may want to rename it though.

src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs
```diff
@@ -117,6 +124,115 @@ public async Task<IActionResult> Test2(
...
+
+        return new EmptyResult();
+    }
+
+    /// <summary>
+    /// Manually trigger ingestion of a SIRE organization event.
+    /// </summary>
+    [HttpPost("sire/ingest")]
+    public async Task<IActionResult> IngestSireOrganization(
+        [FromBody] IngestSireEventCommand command,
```
Alxandr: What is this `IngestSireEventCommand`? Where is it handled?

src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs
```diff
@@ -117,6 +124,115 @@ public async Task<IActionResult> Test2(
...
+    {
+        await sender.Send(command, cancellationToken);
+
+        return Ok(command);
+    }
+
+    /// <summary>
+    /// Manually trigger SIRE enrichment for an organization.
+    /// </summary>
+    [HttpPost("sire/enrich/{orgNo}")]
```
Alxandr: This claims to manually trigger enrichment, but rather it just fetches the org. It should be `sire/org/{orgNo}`.

src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs
acn-dgopa: that was my initial implementation based on the issue but ended up implementing as enricher. i will remove this test method

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
acn-dgopa: After the meeting with skatt, the identifiertype we must be worried is "TaxIdentificationNumber" and this can be either a norewgian org number or person number

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/SireClient.cs
acn-dgopa: I copied it from nprclient but you think the actual json value , will be logged as part of the error message? Also in org document, the person identifier is part of business relation

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
acn-dgopa: I have to dig deep into this as i'm not seeing this in SIRE model

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireBusinessRelationship.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    /// Gets the validity start timestamp.
+    /// </summary>
+    public required DateTimeOffset? ValidFrom { get; init; }
+
+    /// <summary>
+    /// Gets the validity end timestamp.
+    /// </summary>
+    public required DateTimeOffset? ValidTo { get; init; }
```
Alxandr: Register does not support any sort of "valid to" on roles. This should just be removed from here.

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireBusinessRelationship.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    /// Gets the related organization identifier, if the related party is an organization.
+    /// </summary>
+    public required OrganizationIdentifier? RelatedOrganizationIdentifier { get; init; }
+
+    /// <summary>
+    /// Gets the validity start timestamp.
+    /// </summary>
+    public required DateTimeOffset? ValidFrom { get; init; }
```
Alxandr: Register does not support any sort of "valid from" on roles. We could use this as "added on" or "updated on" - but we only have that for the whole party, and not from the source (at the moment). For now, it's cleaner just to remove this.

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireOrganization.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public required string Name { get; init; }
+
+    /// <summary>
+    /// Gets the organization form (e.g. "AS", "ANS"). Stored as free text.
+    /// </summary>
+    public required string UnitType { get; init; }
+
+    /// <summary>
+    /// Gets the unit status. Set to "slettet" when the organization has been deleted.
```
Alxandr: This should not be nullable, and should be (documented) to either `E` (for non-deleted) or `S` (for deleted) organizations.

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireOrganization.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public required bool IsDeleted { get; init; }
+
+    /// <summary>
+    /// Gets the mailing address, or null if no address is available.
+    /// </summary>
+    public required MailingAddressRecord? MailingAddress { get; init; }
+
+    /// <summary>
+    /// Gets the last updated timestamp from the postal address.
```
Alxandr, line 45: From the postal address?

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/IdentifierType.cs
```diff
@@ -0,0 +1,17 @@
+using System.Text.Json.Serialization;
+using Altinn.Authorization.ModelUtils;
+
+namespace Altinn.Register.Integrations.Sire.Organization;
+
+/// <summary>
+/// Specifies the source of a person-party.
+/// </summary>
+[StringEnumConverter]
+public enum IdentifierType
```
Alxandr: I can't see this enum being used? Should it just be deleted?

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/BusinessRelationship.cs
```diff
@@ -0,0 +1,39 @@
+using System.Text.Json.Serialization;
+
+namespace Altinn.Register.Integrations.Sire.Organization;
+
+/// <summary>
+/// Represents a business relationship from the SIRE API 
+/// </summary>
+public sealed record BusinessRelationship
```
Alxandr: Most (if not all) types in `Altinn.Register.Integrations.Sire.Organization` can be `internal`.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocument.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// Gets the date the company was deleted, if applicable.
+    /// </summary>
+    [JsonPropertyName("slettetdato")]
+    public string? DeletedDate { get; init; }
+
+    /// <summary>
+    /// Gets the tax liability type (e.g. "selskap", "personligSkattepliktig").
+    /// </summary>
+    [JsonPropertyName("typeSkattepliktig")]
+    public string? TaxLiabilityType { get; init; }
```
Alxandr: If we don't use a property here, just get rid of it.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+        _lookup = lookup;
+        _timeProvider = timeProvider;
+    }
+
+    /// <summary>
+    /// Returns true when the validTo indicates the
+    /// entity is still currently valid as of now means
+    /// "no termination set, still valid". A non-null value in the past means expired.
+    /// </summary>
+    private static bool IsStillValid(DateTimeOffset? validTo, DateTimeOffset now)
```
Alxandr: We can't deal with values that "will" expire. Instead of this logic, we should give an error if we receive a `validTo` that's more than a few minutes (10 maybe?) into the future. Basically, we should generally treat `validTo is not null` as "no longer valid", but we can add a sanity check.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+            context.AddChildProblem(StdValidationErrors.Required, "/identifikator");
+        }
+        else if (!OrganizationIdentifier.TryParse(input.Identifier, provider: null, out var parsedOrgId))
+        {
+            context.AddChildProblem(ValidationErrors.InvalidOrganizationNumber, "/identifikator");
+        }
+        else
+        {
+            orgId = parsedOrgId;
+        }
```
Alxandr, line 93: Better to reuse existing functionality:

```suggestion
        OrganizationIdentifier? orgId = null;
        if (string.IsNullOrWhiteSpace(input.Identifier))
        {
            context.AddChildProblem(StdValidationErrors.Required, "/identifikator");
        }
        else
        {
		        context.TryValidateChild(
		            path: "/identifikator",
		            input.Identifier,
		            default(OrganizationIdentifierValidator),
		            out orgId);
        }
```

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+
+        // Validate business relationships up-front so any per-relationship problems
+        // are added to the context before we decide whether to bail out.
+        var businessRelationships = MapBusinessRelationships(
```
Alxandr, line 126: This should preferrably be a `context.TryValidateChild` call - though it's not terribly important and can be improved later.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+    private readonly ILocationLookup _lookup;
+    private readonly TimeProvider _timeProvider;
+
+    /// <summary>
+    /// Initializes a new instance of the OrganizationDocumentValidator class using the specified location lookup
+    /// service and time source.
+    /// </summary>
+    /// <param name="lookup">The location lookup service used to validate organization document locations.</param>
+    /// <param name="timeProvider">The clock used to filter out expired postadresse and virksomhetsrelasjon entries.</param>
+    public OrganizationDocumentValidator(ILocationLookup lookup, TimeProvider timeProvider)
```
Alxandr: The validator isn't really a "service" per say. Instead of accepting a TimeProvider here, what's really needed is the "now" time. So just take in a `DateTimeOffset now` instead. That means all the dates in one document is validated using the same "now" instead of being based on how fast we loop.

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireOrganization.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// <summary>
+    /// Gets the unit status. Set to "slettet" when the organization has been deleted.
+    /// </summary>
+    public required string? UnitStatus { get; init; }
+
+    /// <summary>
+    /// Gets whether the organization is deleted.
+    /// </summary>
+    public required bool IsDeleted { get; init; }
```
Alxandr: This can be replaced with `public required DateOnly? DeletedAt { get; init; }`

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+
+        if (context.HasErrors)
+        {
+            validated = null;
+            return false;
+        }
+
+        Debug.Assert(orgId is not null);
+
+        bool isDeleted = !string.IsNullOrWhiteSpace(input.DeletedDate);
```
Alxandr: The date needs to be validated. Not just checked that it's not `null`.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+        MailingAddressRecord? mailingAddress = NormalizeAddress(input.PostalAddress);
+
+        validated = new SireOrganization
+        {
+            OrganizationIdentifier = orgId!,
+            Name = input.CompanyName,
```
Alxandr, line 146: `input.CompanyName` needs to be validated. Should produce `StdValidationErrors.Required` if it's missing (or empty).

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+        var lines = NormalizeAddressLines(address.AddressLines, isForeign: false);
+        var postalCode = NormalizeNorwegianPostalCode(address.PostalCode);
+        var city = address.City?.Trim();
+
+        if (lines.Count == 0 && postalCode is null && city is null)
+        {
+            return null;
+        }
+
+        return new MailingAddressRecord
```
Alxandr, line 212: For both FREG and ER we have logic that adds the postal-code/city to the address lines (if it's not already present). We should probably do the same for SIRE.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        if (lines.Count == 0 && postalCode is null && city is null)
+        {
+            return null;
+        }
+
+        return new MailingAddressRecord
+        {
+            Address = lines.Count > 0 ? string.Join(' ', lines) : null,
```
Alxandr, line 249: For both FREG and ER we have logic that adds the postal-code/city to the address lines (if it's not already present). We should probably do the same for SIRE.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        foreach (var line in addressLines)
+        {
+            if (string.IsNullOrWhiteSpace(line))
+            {
+                continue;
+            }
+
+            var normalized = RemoveCareOfPrefix(line);
```
Alxandr: Should this be done for each line?

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+            }
+
+            var relatedId = rel.RelatedIdentifier;
+            if (relatedId is null || string.IsNullOrWhiteSpace(relatedId.Value))
+            {
+                context.AddChildProblem(StdValidationErrors.Required, path + "/relatertIdentifikator");
+                continue;
+            }
+
+            if (!string.Equals(relatedId.IdentifierType, "taxIdentificationNumber", StringComparison.Ordinal))
```
Alxandr: What if it's an org number?

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,394 @@
...
+            {
+                context.AddChildProblem(
+                    ValidationErrors.InvalidValue,
+                    path + "/relatertIdentifikator/identifikatortype");
+                continue;
+            }
+
+            PersonIdentifier? personId = null;
+            OrganizationIdentifier? orgId = null;
+            if (PersonIdentifier.TryParse(relatedId.Value, provider: null, out var parsedPerson))
```
Alxandr: Shouldn't this be controlled by the `identifikatortype`? If it should, use the `PersonIdentifierValidator` and `OrganizationIdentifierValidator` to validate.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/SireRoleMapper.cs
```diff
@@ -0,0 +1,84 @@
...
+/// </remarks>
+public static class SireRoleMapper
+{
+    /// <summary>
+    /// Maps a SIRE <c>relasjonstype</c> to the corresponding Altinn role identifier.
+    /// </summary>
+    /// <param name="relasjonstype">The technical name from SIRE (e.g. "komplementar").</param>
+    /// <returns>The Altinn role identifier (e.g. "komplementar", "daglig-leder").</returns>
+    /// <exception cref="ArgumentException">Thrown when the relasjonstype is not recognized.</exception>
+    public static string GetRoleIdentifier(string relasjonstype) => relasjonstype switch
```
Alxandr: It would be much better if this was implemented as a `IValidator<string, string>`. Similarly, `SireOrganizationFormMapper` can be a `IValidator<string?, string>` (notice the difference in nullability). This matters if we receive multiple values we don't recognice in 1 response, because instead of throwing we aggregate up the errors. We also get information on where in the document the error originated.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Altinn.Register.Integrations.Sire.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
```
Alxandr: This block can be removed.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/SireClient.cs
```diff
@@ -0,0 +1,102 @@
...
+        _client = client;
+        _lookupProvider = lookupProvider;
+        _timeProvider = timeProvider;
+    }
+
+    /// <inheritdoc/>
+    public async Task<Result<SireOrganization>> GetOrganization(OrganizationIdentifier organizationIdentifier, CancellationToken cancellationToken)
+    {
+        using var response = await _client.GetAsync(
+            $"v1/digdir/{Uri.EscapeDataString(organizationIdentifier.ToString())}",
```
Alxandr: `.ToString()` can be removed.

src/apps/Altinn.Register/src/Altinn.Register.Persistence/Migration/v0.55-add-organization-source-srs/00-org-source-enum.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TYPE register.organization_source RENAME VALUE 'sdf' TO 'srs';
+ALTER TYPE register.external_role_source ADD VALUE 'srs'; -- RegisteredWithSkatteetaten (skatteetatens register av indre selskap)
```
Alxandr: SIRE is not a role source.

src/apps/Altinn.Register/src/Altinn.Register/Controllers/DebugController.cs
```diff
@@ -117,6 +119,103 @@ public async Task<IActionResult> Test2(
...
+        HttpProxyResult.CopyHeaders(response.Content.Headers, Response);
+        await response.Content.CopyToAsync(Response.Body, cancellationToken);
+
+        return new EmptyResult();
+    }
+
+    /// <summary>
+    /// Manually trigger SIRE enrichment for an organization.
+    /// </summary>
+    [HttpPost("sire/org/{orgNo}")]
```
Alxandr, line 192: Missing `[Authorize(Policy = "Debug")]`

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,127 @@
...
+    {
+        Debug.Assert(context.Party is OrganizationRecord);
+        Debug.Assert(context.Party is { OrganizationIdentifier.HasValue: true });
+
+        var result = await _sireClient.GetOrganization(context.Party.OrganizationIdentifier.Value, cancellationToken);
+
+        if (result.IsProblem && result.Problem.ErrorCode == Problems.OrganizationNotFound.ErrorCode)
+        {
+            // Organization not found in SIRE — mark as deleted
+            context.Party = ((OrganizationRecord)context.Party) with
```
Alxandr: I think this is a bad idea. I know I commented that just returning here is wrong - but I think it's less of an issue than I originally thought. We will have to improve this later, but it requires some changes to the `Enricher` contract, so that can come later.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,127 @@
...
+            IsDeleted = organization.IsDeleted,
+            Source = OrganizationSource.RegisteredWithSkatteetaten,
+            ModifiedAt = FieldValue.From(organization.LastUpdated),
+            MailingAddress = organization.MailingAddress,
+        };
+
+        await MapRoleAssignments(context, organization, cancellationToken);
+    }
+
+    private async Task MapRoleAssignments(
```
Alxandr: Does this need to be async? If not, you can remove the cancellation token. If it does, the token should likely be used.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,127 @@
...
+        A2PartyImportSagaEnrichmentRunContext context,
+        SireOrganization organization,
+        CancellationToken cancellationToken)
+    {
+        var relationships = organization.BusinessRelationships;
+
+        if (relationships is not { Length: > 0 })
+        {
+            context.RoleAssignments[ExternalRoleSource.RegisteredWithSkatteetaten] = PartyExternalRoleAssignmentsUpdate.Full.Empty
+;
```
Alxandr: Semicolon fell off the previous line 😛

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,127 @@
...
+                    OrganizationIdentifier = orgId,
+                };
+            }
+            else
+            {
+                // Unreachable: OrganizationDocumentValidator rejects any SIRE relationship
+                // that doesn't carry exactly one of RelatedPersonIdentifier or
+                // RelatedOrganizationIdentifier. Asserting here turns drift in that contract
+                // into a loud failure instead of a silent data loss.
+                throw new InvalidOperationException(
```
Alxandr: Preffer `ThrowHelper.ThrowInvalidOperationException`.

src/apps/Altinn.Register/src/Altinn.Register.Core/Sire/SireOrganization.cs
acn-dgopa: That was just added initially. There is no updated timestamp for the whole document. only the address and virksomhetsrelasjon that gets the timestamps. Otherwise they have a date the org was established. So i gues we need to use the timestamp at the time of insert

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/IdentifierType.cs
acn-dgopa: yes, crated it but dont see a need for that as its always 'taxidentificationnumber' that we will be receiving

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
acn-dgopa: I've sent Skatt a clarifying question about whether opphoerstidspunkt is historical-only vs. can carry future-dated planned terminations, and whether there's an erGjeldende-style flag. Current design assumes historical-only with a sanity-check error for future dates.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
acn-dgopa: the idenfifiertype is the same for both person and orgnumber and only type that we need to look for in SIRE documents according to the skatt.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
acn-dgopa: the idenfifiertype is the same for both person and orgnumber and only type that we need to look for in SIRE documents according to the skatt.

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
acn-dgopa: ok, leaving this for now and update later

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/SireClient.cs
```diff
@@ -0,0 +1,106 @@
...
+        _lookupProvider = lookupProvider;
+        _timeProvider = timeProvider;
+    }
+
+    /// <inheritdoc/>
+    public async Task<Result<SireOrganization>> GetOrganization(OrganizationIdentifier organizationIdentifier, CancellationToken cancellationToken)
+    {
+        using var response = await _client.GetAsync(
+            $"v1/digdir/{organizationIdentifier}",
+            cancellationToken);
```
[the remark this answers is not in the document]
Alxandr: Only constructable from user-input in debug.

src/apps/Altinn.Register/src/Altinn.Register.Integrations.Sire/Organization/OrganizationDocumentValidator.cs
```diff
@@ -0,0 +1,501 @@
...
+    /// </summary>
+    private static void ValidateNotFarFuture(
+        ref ValidationContext context,
+        DateTimeOffset? date,
+        DateTimeOffset now,
+        string path)
+    {
+        if (date is { } at && at > now + FutureDateGrace)
+        {
+            context.AddChildProblem(ValidationErrors.InvalidValue, path);
```
Alxandr: This should have a `detail` argument (`context.AddChildProblem(ValidationErrors.InvalidValue, path, detail: "The provided expiry '{at}' is more than 10 minutes into the future");`)

src/apps/Altinn.Register/src/Altinn.Register/PartyImport/A2/Enrichers/SireEnricher.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        context.Party = ((OrganizationRecord)context.Party) with
+        {
+            DisplayName = organization.Name,
+            UnitType = organization.UnitType,
+            UnitStatus = organization.UnitStatus,
+            IsDeleted = organization.DeletedAt is not null,
+            DeletedAt = FieldValue.From(organization.DeletedAt),
+            Source = OrganizationSource.RegisteredWithSkatteetaten,
+            ModifiedAt = FieldValue.From(organization.LastUpdated),
```
Alxandr: ModifiedAt should not be set from SIRE. This is the date at which it was modified *in altinn*. Not at the source. We don't have a field for that yet.

Alxandr (approved):

